PR #29, opened 2012-12-20
Robust hash ring failure retry mechanism

charsyam:
There are 2 step to check connection.

step 1] trying to connect in server_failure

  If it succeed,  going to step 2

step 2] trying to send heartbeat command to fail server

 If it succeed, it will update hash ring.

and it uses array just to connect failed servers to prevent infinity loop for connecting

and in heartbeat stage. It uses msg_two_insert for timeout

but, if there is no timeout setting in conf, It will just wait for response like other message.

### DIFF
--- a/src/hashkit/nc_ketama.c
+++ b/src/hashkit/nc_ketama.c
@@ -90,12 +90,8 @@ ketama_update(struct server_pool *pool)
         struct server *server = array_get(&pool->server, server_index);
 
         if (pool->auto_eject_hosts) {
-            if (server->next_retry <= now) {
-                server->next_retry = 0LL;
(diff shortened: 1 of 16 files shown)


manjuraj:
There are too many threads going on in this pull request. However for @vinay-ganesh and @hrishimantri  the following changes fixes the issue of redis server not getting ejected when it is unresponsive and loading dataset from disk:
* https://github.com/twitter/twemproxy/commit/ef453130e321974b332dd99d585ae7285eee4b5d
* https://github.com/twitter/twemproxy/commit/5e082b8d3ae0479b8f019a5f53e1e289f66ddfe9

PR #123, opened 2013-06-13
Use sysconfdir & localstatedir for conf/pid/logs

paravoid:
Use autoconf-supplied paths as the default values for the configuration
file, the pidfile and the logfile. So, for example:
  /etc/nutcracker/nutcracker.yml
  /var/run/nutcracker/nutcracker.pid
  /var/log/nutcracker/nutcracker.log

The separate subdirectory under /var/run was picked since nutcracker
doesn't possess the capability of dropping privileges by itself, so we
need to have a system user writeable directory to place the pidfile.

Similarly, we need a separate /var/log subdir as logs can't go to
syslog, and out of convenience (need to be able to perform logrotate,
cleanup upon removing a package etc.)

The separate subdirectory under /etc/ is just for being pretty
(top-level /etc conffiles are ugly) and for future use (multiple configs
for the same daemon?)

All of the above can of course be used in the case where a single init
script spawns multiple nutcrackers, each with a separate config, pidfile
and logfile, for example for extra redundancy reasons.

(also fix an underlying bug where NC_PID_FILE was completely unused)

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,4 +1,11 @@
 MAINTAINERCLEANFILES = Makefile.in
+BUILT_SOURCES = cfg-options.h
+CLEANFILES = cfg-options.h
+
+cfg-options.h: Makefile
+	$(AM_V_GEN)echo '#define NC_CONF_PATH "$(sysconfdir)/nutcracker/nutcracker.yml"' >$@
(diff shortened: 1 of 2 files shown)


paravoid:
Ping?

paravoid:
This is a 6-year old unacknowledged patch which I've given up on. If someone is interested in reviving this, I can be persuaded to sign a CLA, but otherwise I have no intentions of bothering with that as things stand. HTH!

PR #177, opened 2013-12-05
[New Feature] adding to always resolve hostname with always_host_resolve option.

charsyam:
if set always_host_resolve to true:

It always tries to resolve when reconnect to server.

before patch
1. read conf
2. resolve host address
3. connect
4. disconnect
5. connect with 2's info

after patch("always_host_resolve: true)
1. read conf
2. resolve host address
3. connect
4. disconnect
5. resolve host address again.
6. connect with 5's info

### DIFF
--- a/src/nc_conf.c
+++ b/src/nc_conf.c
@@ -74,6 +74,10 @@ static struct command conf_commands[] = {
       conf_set_bool,
       offsetof(struct conf_pool, redis) },
 
+    { string("always_host_resolve"),
+      conf_set_bool,
+      offsetof(struct conf_pool, always_host_resolve) },
+
     { string("preconnect"),
       conf_set_bool,
       offsetof(struct conf_pool, preconnect) },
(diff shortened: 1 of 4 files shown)


tom-dalton-fanduel:
@manjuraj I'd be quite interested in this feature, as it would allow twemproxy to work more seamlessly with Elasticache in the case of a master->slave failover, or if a slave instance fails and is replaced with new hardware - in both cases the DNS entry for the primary/slave endpoint gets updated but as things stand we'd need a separate process watching for that to happen and restarting twemproxy accordingly.

If you think the code looks OK from a happy-to-merge point of view, I'd be happy to give this some testing to make sure it's still working as expected.

manjuraj:
@tom-dalton-fanduel -- https://github.com/twitter/twemproxy/issues/355#issuecomment-113575050.

manjuraj:
fixed with this patch:  https://github.com/twitter/twemproxy/commit/fd34c29bb69861b37eea786bd84a7e437ab8bb7c

tom-dalton-fanduel:
Fantastic!

PR #178, opened 2013-12-09
Fix bug: crashing when there is no lived server in heartbeat branch

charsyam:
I found one important bug. and fixed it. Thank you.

### DIFF
--- a/src/nc_server.c
+++ b/src/nc_server.c
@@ -290,12 +290,13 @@ server_failure(struct context *ctx, struct server *server)
 
     server->failure_count = 0;
 
+    add_failed_server(ctx, server);
+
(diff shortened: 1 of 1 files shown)


src/nc_server.c
```diff
@@ -290,12 +290,13 @@ server_failure(struct context *ctx, struct server *server)
 
     server->failure_count = 0;
 
+    add_failed_server(ctx, server);
```
TysonAndre, line 293: LGTM - I'd independently run into the same bug and had done something similar as a part of 82d4bba1d64a71f533e4b5886d3a687e459451ba in a fork before I'd seen this when implementing support for recursive failover

Issue #158, opened 2013-10-28
mbuf_split() splits by copying data in quadratic instead of linear time


neeleshkorade:
Hi All

We are looking to use TwemProxy with Redis for sharding. We have use cases where we may need to fetch about 10k keys in one go from across multiple shards. However, when I try this with TwemProxy on a test setup (described below), it takes about 1.7 seconds to return. If I fired the same request on a single Redis instance directly, it returns in about 16ms.

Would someone know if I have missed out some obvious performance tuning option? Any ideas on what we could do to better this number? We need mget response time to be within sub 300ms range.

Note that the hosts are shared with other processes (and not dedicated for my performance tests). I am giving the details below to give an indication of the leftover capacity.

Setup Details-
Host One- 8 core, 3.3 GHz (5% utilization during test run), 16GB RAM (6GB available) hosts two redis instances and a TwemProxy instance

Host Two- 8 core, 3.3 GHz (5% utilization during test run), 16GB RAM (9.5GB available) hosts two redis instances

TwemProxy config- 
hash: murmur
distribution: ketama
servers:
- host1:6379:1
- host1:6380:1
- host2:6379:1
- host2:6380:1

Thank you

TysonAndre:
> For example, if my input buffer from read syscall contains 10 messages = [1, 2, 3, 4, 5, 6, ... 10], we leave existing message "1" in its current mbuf and copy messages from [2,3,4,5, ...10] to a new mbuf. Once message "1" is processed, we then we copy messages from [3,4,5,6,...10] to a new mbuf and so on and on. So, to split messages [1,2,3...10] across 10 mbufs we are doing quadratic instead of linear copies. This is really unfortunate, and I think if we fix this we twemproxy will consume less cpu. The workaround for now is that for scenarios where you need high throughput and low cpu util %, you use -mbuf-size=512.

If I'm reading this right, this is only a performance issue with pipelined commands (and common redis clients make pipelining convenient)

I'd wonder if it'd make sense to conditionally swap the pointers with the "new" mbuf if mbuf_split had a larger buffer for a new message (`nmsg`) - that would hopefully make it take less time.

(i.e. replace the old mbuf's pointers with the smaller copy at the beginning, and use the old pointers with the larger size for the new mbuf)

PR #210, opened 2014-03-31
MGET improve

idning:
hi, @manjuraj, I'm trying to import mget performance by rewrite it like this::

orig: 

```
mget k1 k2 k3 k4 k5 k6 k7 k8 k9 
```

after rewrite:

```
mget k1 k3 k6
mget k2 k4 k7
mget k5 k8 k9
```

this will cost less time on `req_done()` function calls, 

I use this redis-benchmark::

```
    if (test_is_selected("mget")) { 
#define N 1000
        const char *argv[N+1];
        argv[0] = "MGET";
        for (i = 1; i < N+1; i += 1) {
            argv[i] = "key:__rand_int__";
        }
        len = redisFormatCommandArgv(&cmd,N+1,argv,NULL);
        char tmp[1024];
        sprintf(tmp, "MGET (%d keys)", N);
        benchmark(tmp, cmd,len);
        free(cmd);
    }
```

here is the  benchmark result  on my laptop:

1000 keys mget::

```
ning@ning-laptop:~/idning-github/redis-mgr$ redis-benchmark -n 1000 -t mget -p 4001 -r 100000000
====== MGET (1000 keys) ======
  1000 requests completed in 21.66 seconds
  50 parallel clients
  3 bytes payload
  keep alive: 1

100.00% <= 1268 milliseconds
46.18 requests per second

ning@ning-laptop:~/idning-github/redis-mgr$ redis-benchmark -n 1000 -t mget -p 4000 -r 100000000
====== MGET (1000 keys) ======
  1000 requests completed in 1.64 seconds
  50 parallel clients
  3 bytes payload
  keep alive: 1

100.00% <= 102 milliseconds
611.25 requests per second  (improved )

(benchmark against redis got 1292 requests per second)
```

100keys mget::

```
ning@ning-laptop:~/idning-github/redis-mgr$ redis-benchmark -n 10000 -t mget -p 4001 -r 100000000
====== MGET (100 keys) ======
  10000 requests completed in 8.57 seconds
  50 parallel clients
  3 bytes payload
  keep alive: 1

100.00% <= 56 milliseconds
1166.59 requests per second

ning@ning-laptop:~/idning-github/redis-mgr$ redis-benchmark -n 10000 -t mget -p 4000 -r 100000000
====== MGET (100 keys) ======
  10000 requests completed in 1.88 seconds
  50 parallel clients
  3 bytes payload
  keep alive: 1

100.00% <= 17 milliseconds
5316.32 requests per second  (improved )

(benchmark against redis got 11668.61 requests per second)
```

the performance is about `5x-10x` (about 1/2 of redis instance) 
@neeleshkorade  will be happy to see this result (https://github.com/twitter/twemproxy/issues/158)

I need a detail benchmark later. and I need your advise.

### DIFF
--- a/src/nc_log.c
+++ b/src/nc_log.c
@@ -133,11 +133,10 @@ _log(const char *file, int line, int panic, const char *fmt, ...)
 {
     struct logger *l = &logger;
     int len, size, errno_save;
-    char buf[LOG_MAX_LEN], *timestr;
+    char buf[LOG_MAX_LEN];
     va_list args;
-    struct tm *local;
-    time_t t;
     ssize_t n;
+    struct timeval tv;
 
     if (l->fd < 0) {
         return;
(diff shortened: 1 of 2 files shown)


thinkingfish:
The behavior of MGET return order is not specified in the protocol, so I would consider it implementation dependent (on the client).

Issue #213, opened 2014-04-04
auto_eject_drop as an alternative to auto_eject_hosts


mezzatto:
Background:
- modula distribution
- using twemproxy in front of redis servers that are acting as databases, not as caches
- data is sharded between these redis servers
- master / slave configuration
- application is 99% reads, almost no writes
- client balances the load between the masters and the slaves instances. For every request the client chooses if it goes to the master or to the slave pool in twemproxy (50/50 weight)

If one or more redis server goes offline and I choose the twemproxy pool that has the failing redis server and the key is hashed to this redis server, I have to wait the timeout before I get my reply and try the other twemproxy pool. Every request that goes to this redis server has to wait. This is bad... and I cannot use auto_eject_hosts since it changes the shards / hash ring.

I though about implementing a new config, called `auto_eject_drop` with values `true` (current behavior, default) and `false` (my proposal) that lets you:
- when a server fails, it wont decrement the live servers of the continuum (this way the shards wont change)
- immediately replies with "ERR timeout" if some key tries to ask a failing redis server

The way to do that is maybe changing req_forward() so that if a faling server got asked, the message will not be enqueued and. I dont know if this is the best solution but is the one that has minimal code change.

Any thoughts?

TysonAndre:
Not exactly the same as your use case, but see https://github.com/twitter/twemproxy/issues/608

An upcoming 0.6.0 is planned with the following:
- Support for redis sentinel, redis's official way for host failover
- A `failover` pool for memcache - instead of ejecting hosts or attempting to reconnect to failing hosts, send requests for keys corresponding to failing hosts to the failover pool instead of the failing host in the main pool

PR #214, opened 2014-04-04
New option "auto_eject_drop"

mezzatto:
A boolean value that controls if auto ejected hosts should be dropped from the hash ring. If set to false, failing hosts will immediately reply timeout. Defaults to true.

See https://github.com/twitter/twemproxy/issues/213 for more information

### DIFF
--- a/src/nc_message.c
+++ b/src/nc_message.c
@@ -145,7 +145,6 @@ void
 msg_tmo_insert(struct msg *msg, struct conn *conn)
 {
     struct rbnode *node;
-    struct server *server;
     int timeout;
 
     ASSERT(msg->request);
@@ -154,13 +153,6 @@ msg_tmo_insert(struct msg *msg, struct conn *conn)
     timeout = server_timeout(conn);
     if (timeout <= 0) {
         return;
-    }
-
-    server = conn->owner;
-
-    /* insert already expired */
-    if (server->dead) {
-        timeout = -1;
     }
 
     node = &msg->tmo_rbe;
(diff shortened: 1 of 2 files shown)


pauldavidgilligan-msm:
:+1: definitely a good one! nice job.

codekitchen:
Am I reading this change correctly that `auto_eject_drop` is only implemented in the modula distribution? That seems like a pretty big caveat.

mezzatto:
@codekitchen, that's right.

In ketama distribution, this option doesn't seem to make sense (or does it?).

In random distribution, the implementation is indeed not done yet (but it's trivial to get done). I had no practical application for it when I first implemented the original code, so I though that the less code I touched the faster the pull would be accepted... :)

codekitchen:
Heh well, I can't speak for the maintainers but I would certainly want the limitation to at least be documented.

I'm curious why you say it doesn't make sense for ketama -- to venture a guess, I think you're saying that why would you be using consistent hashing if you don't plan on modifying the ring on failure? In our case, it's because we want consistent hashing for when we explicitly resize the ring, but we also prefer this behavior of not modifying on node failure, rather just failing fast.

mezzatto:
@manjuraj, would you be willing to accept this pull If I documented that this feature is only for the modula distribution? Or do you think that, for completeness, it would be for the better to make the changes in ketama distribution as @codekitchen rised some valid use cases?

mezzatto:
After all theses years, revisiting this PR made me realize that it's a https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern

mezzatto:
@CLAassistant you are a bot robot. Of course I'm a github user :D

PR #257, opened 2014-09-04
hot-reload

idning:
how it works:
1. when we receive USR1 signale, we will fork and exec with new binary and new config.
2. the new process will inherited all listen-socket from the old process, including:
   - stat socket
   - listen socket for each pool
3. after the new process is already running, the old process will close all listen socket.
   and wait for 3 seconds (for those who has already connected with old process)
   3 seconds later, it close all client socket and shut down
   
   (however, the keep-alive connectin will be closed, we can not wait forever)

TODO:
- make it work for `unix domain socket`
- need test on nc_kqueue

test cases:

https://github.com/idning/test-twemproxy/blob/master/test_system/test_reload.py

### DIFF
--- a/src/event/nc_epoll.c
+++ b/src/event/nc_epoll.c
@@ -36,6 +36,16 @@ event_base_create(int nevent, event_cb_t cb)
         return NULL;
     }
 
+    status = fcntl(ep, F_SETFD, FD_CLOEXEC);
+    if (status < 0) {
+        log_error("fcntl for FD_CLOEXEC on e %d failed: %s", ep, strerror(errno));
+        status = close(ep);
+        if (status < 0) {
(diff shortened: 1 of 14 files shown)


mckelvin:
:+1: The feature is great, but may not be friendly enough with Docker as the original process is killed any way. What about using inotify([mcrouter is using it](https://github.com/facebook/mcrouter/blob/master/mcrouter/FileDataProvider.cpp#L13))?

mckelvin:
An issue: if nutcracker is run as a daemon, after hot-reload(`kill -10`) the pid file is not updated.

vlm:
Folks, I created an alternative patch which does not fork the process. It supports the unix sockets, and does not drop the connection. Hope that too can be valuable. Will create a pull request by the end of this week.

PR #283, opened 2014-11-27
fix delete parsing for memcache

charsyam:
I think we should support memcache delete argument. because memcached delete spec allow arguments.

See process_delete_command in https://github.com/memcached/memcached/blame/master/memcached.c 

``` c
    if (ntokens > 3) {
        bool hold_is_zero = strcmp(tokens[KEY_TOKEN+1].value, "0") == 0;
        bool sets_noreply = set_noreply_maybe(c, tokens, ntokens);
        bool valid = (ntokens == 4 && (hold_is_zero || sets_noreply))
            || (ntokens == 5 && hold_is_zero && sets_noreply);
        if (!valid) {
            out_string(c, "CLIENT_ERROR bad command line format.  "
                       "Usage: delete <key> [noreply]");
            return;
        }
    }
```

 so delete command syntax like below:

``` c
delete <key>
delete <key> 0
delete <key> <numbers> noreply
```

Actually, memcached doesn't handle <numbers> but, it doesn't return ERROR, just it ignores internally.
so. I just follow current memecached behavior. :)

### DIFF
--- a/src/proto/nc_memcache.c
+++ b/src/proto/nc_memcache.c
@@ -127,6 +127,8 @@ memcache_parse_req(struct msg *r)
     struct mbuf *b;
     uint8_t *p, *m;
     uint8_t ch;
+    uint8_t *last_num;
+    int last_num_size;
     enum {
         SW_START,
         SW_REQ_TYPE,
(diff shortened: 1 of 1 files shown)


rohitpaulk:
Note: The usage in python-memcached was fixed in https://github.com/linsomniac/python-memcached/pull/27

PR #300, opened 2014-12-22
add note for python memcached

charsyam:
adding some comments for usage of python-memcacahed library.
adding How to solve problem.


deep011:
I think it is better to support "delete key 0/r/n" for twemproxy, because memcached support this command.

PR #309, opened 2014-12-31
Redis key-value distribution with twemproxy compatible with phpredis redisArray

amolrajoba:
Implementation to use key-value distribution with twemproxy compatible with phpredis redisArray (https://github.com/phpredis/phpredis/blob/master/arrays.markdown)

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,6 @@ For example, the configuration file in [conf/nutcracker.yml](conf/nutcracker.yml
        - 127.0.0.1:11214:100000
        - 127.0.0.1:11215:1
 
-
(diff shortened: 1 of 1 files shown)


amolrajoba:
nicolasff can you please tell me why this pull request is still not accepted & merged?

amolrajoba:
rcrc32 (implementation compatible with phpredis [redisArray](https://github.com/phpredis/phpredis/blob/master/arrays.markdown) )


If there is a need to use key-value distribution as per phpredis [redisArray](https://github.com/phpredis/phpredis/blob/master/arrays.markdown) use hashing function 'rcrc32' and distribution mode 'redisarray'.

PR #313, opened 2015-01-15
hashkit: Add correct fnv1* implementations.

mckelvin:
Old fnv1\* implementations(fnv1_64/fnv1a_64/fnv1_32/fnv1a_32) are not correct as per spec but are compatible with libmemcachd. The corresponding correct implementations are fnv1_64a/fnv1a_64a/fnv1_32a/fnv1a_32a.

~~This issue is similar to 653b05f861ae3ce0ff143e1067af378f677a4d29 (@manjuraj).~~

http://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function says:

```
The XOR is an 8-bit operation that modifies only the lower 8-bits of the hash value.
```

Suppose key is a utf-8 encoded string `"\xe4\xb8\xad"`(Chinese Character`中` in UTF-8). Let's take `\xe4`(`0xe4 >= 0b10000000`) as an example:

``` c
(uint32_t)(char)0xe4 = 4294967268 // buggy
(uint32_t)(uint8_t)(char)0xe4 = 228 // correct
(uint32_t)(unsigned char)(char)0xe4 = 228 // also correct
```

cc @windreamer

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,10 +86,14 @@ nutcracker can be configured through a YAML file specified by the -c or --conf-f
  + crc16
  + crc32 (crc32 implementation compatible with [libmemcached](http://libmemcached.org/))
  + crc32a (correct crc32 implementation as per the spec)
- + fnv1_64
+ + fnv1_64 (fnv\* implementations compatible with [libmemcached](http://libmemcached.org/), same below)
  + fnv1a_64
  + fnv1_32
  + fnv1a_32
+ + fnv1_64a (correct fnv\* implementations as per the spec, same below)
+ + fnv1a_64a
+ + fnv1_32a
+ + fnv1a_32a
  + hsieh
  + murmur
  + jenkins
--- a/src/hashkit/nc_fnv.c
+++ b/src/hashkit/nc_fnv.c
@@ -22,6 +22,11 @@ static uint64_t FNV_64_PRIME = UINT64_C(0x100000001b3);
 static uint32_t FNV_32_INIT = 2166136261UL;
 static uint32_t FNV_32_PRIME = 16777619;
 
+
+/*
+ * fnv* implementations compatible with libmemcached library. Unfortunately
+ * these implementations do not return fnv* as per spec.
+ */
 uint32_t
 hash_fnv1_64(const char *key, size_t key_length)
 {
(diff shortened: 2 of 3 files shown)


mckelvin:
@manjuraj @idning Would you please review the PR?

idning:
I don't think this make sense. anyone else support it?

mckelvin:
@idning The old implementation is just wrong. In our real production environment, this buggy fnv1a implementation may lead to unbalanced data distribution.

Also in [the fnv home page](http://www.isthe.com/chongo/tech/comp/fnv/), there are some solved **Zero-hash challenges**, which can be used as a sanity test. For example:

```
SOLVED! What is the shortest binary data set for which the 32-bit FNV-1a hash is 0?
32-bit solution by Russ Cox <rcs Email address> on 2011-Mar-02:
The solutions of length 4 are:

FNV32_1a(cc 24 31 c4) = 0
FNV32_1a(e0 4d 9f cb) = 0
```

The old fnv1a_32 implementation will not pass this case.

```c
#include <stdio.h>
#include <stdint.h>

static uint32_t FNV_32_INIT = 2166136261UL;
static uint32_t FNV_32_PRIME = 16777619UL;


uint32_t
hash_fnv1a_32(const char *key, size_t key_length)
{
    uint32_t hash = FNV_32_INIT;
    size_t x;

    for (x= 0; x < key_length; x++) {
      uint32_t val = (uint32_t)key[x];
      hash ^= val;
      hash *= FNV_32_PRIME;
    }

    return hash;
}


uint32_t
hash_fnv1a_32a(const char *key, size_t key_length)
{
    uint32_t hash = FNV_32_INIT;
    size_t x;
    const uint8_t* unsigned_key = (uint8_t*)key;

    for (x= 0; x < key_length; x++) {
      uint32_t val = (uint32_t)unsigned_key[x];
      hash ^= val;
      hash *= FNV_32_PRIME;
    }

    return hash;
}


int main(int argc, const char *argv[])
{
    const char key[] = "\xcc\x24\x31\xc4";
    uint32_t v1 = hash_fnv1a_32(key, 4U);
    uint32_t v2 = hash_fnv1a_32a(key, 4U);
    printf("old: %u, new: %u\n", v1, v2);
    return 0;
}

```

I guess the bug is [originally introduced in libmemcached](https://github.com/mckelvin/libmemcached/blob/master/libhashkit/fnv_32.cc).

mckelvin:
Unittest support is added in #315 .  Please review that first. I'll add unittest for this PR after #315 is merged to make sure the new added implementation is bugfree (hopefully :smile: ).

mckelvin:
@manjuraj  @idning please review.

idning:
we don't need a correct algorithm here, we just need a algorithm which can distribute keys evenly.

adding 4 more hash function may make user more confuse.

mckelvin:
**evenly?** Technically no. [SMHasher](https://code.google.com/p/smhasher/) is a test suite designed to test the distribution, collision, and performance properties of non-cryptographic hash functions. fnv32 implementation in twemproxy won't pass test cases in that.

You can checkout the latest code of SMHasher on svn and apply [the patch](https://gist.github.com/mckelvin/655a1cf51d2bc2568dcc#file-reproduce_fnv_32_bug-patch) based on revision 152 to reproduce.

TheRealBill:
In my opinion we need any algorithm we use, which is externally defined, to return the correct results. If others correctly implement them, our results must match. I'll take consistency over "even" any day. If our results, given the same inputs, are different we should not call the algorithm by the name of one we don't match. 

Cheers,
Bill

idning:
@mckelvin @therealbill sorry for delay such a long time.

in our production environment, fnv1a_64 is even enough, but recently I found that if we want to map a key to a server with python or other language, the buggy algorithms is a problem, so I agree that we need a correct algorithm.

I will merge this ASAP, thanks guys

mckelvin:
@idning :smiley:

sysu2012zzp:
It is good to correct the hash function, but I wonder if we exactly need more hash functions.
Is it possible to revise fnv1* functions in place instead of creating new functions? are there any necessary to be compatible with libmemcachd?

charsyam:
@Marcus366 
If just revise hash functions.
There is a problem, if some one upgrade twemproxy, key hashing will be broken.

mckelvin:
Closing the stale PR.

PR #317, opened 2015-01-28
Fix SORT command to support arguments

areina:
First of all, my apologies because it was me who introduced this bug.

I sent a PR adding support for SORT command but it was wrong and it wasn't accepting arguments (so it was totally useless). This PR solves this issue.

### DIFF
--- a/src/proto/nc_redis.c
+++ b/src/proto/nc_redis.c
@@ -61,7 +61,6 @@ redis_arg0(struct msg *r)
     case MSG_REQ_REDIS_EXISTS:
     case MSG_REQ_REDIS_PERSIST:
     case MSG_REQ_REDIS_PTTL:
-    case MSG_REQ_REDIS_SORT:
     case MSG_REQ_REDIS_TTL:
     case MSG_REQ_REDIS_TYPE:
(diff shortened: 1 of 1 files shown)


charsyam:
@areina @idning @manjuraj 
I'm sorry to say this. but, I think it is better removing to support "sort" command.
Because sort can use some external key, "see this http://redis.io/commands/sort"
so twemproxy can't support sort.

for example
```c
set three 3
set two 2
set one 1
zadd z1 3 three
zadd z1 2 two
zadd z1 1 one
sort z1 ALPHA <-- it's ok
sort z1 by * <-- It will search key one, two, three to get one, two, three
```

charsyam:
@areina @idning @manjuraj 
I think it's ok. just remainning SORT command like now.
It works only like below
```c
 SORT <key>
```

so it works fine now.
But I oppose allowing options for SORT Command

unleashed:
FYI, commit d372c3432ade35388b144e0702bf4e11540b345c did this same change to master, *coding-cowboy* style. :astonished:

manjuraj:
@unleashed good point. closing this pull request as it has already been fixed

PR #318, opened 2015-02-04
remove supporting sort command

charsyam:
Because sort can use some external key, "see this http://redis.io/commands/sort"
so twemproxy can't support sort.

for example

```c
set three 3
set two 2
set one 1
zadd z1 3 three
zadd z1 2 two
zadd z1 1 one
sort z1 ALPHA <-- it's ok
sort z1 by * <-- it's wrong. It will search key one, two, three to get one, two, three
```


charsyam:
But, It's ok just remaining SORT Command like now.
It just works with SORT <key>, so it works fine.

PR #324, opened 2015-03-02
twemproxy work with redis-sentinel

andyqzb:
Hi, @manjuraj @idning , I made a patch for twemproxy to let it work with redis-sentinel.
This patch is implemented as the design mentioned in issue 297.
In addition, I add keepalive to the sentinel pub/sub connection. Because sentinel pub/sub connection don't have heartbeat, we must add keepalive to check if the connection is dead. The implemention of nc_set_keepalive is copied from redis.
You can configure a server_pool like sigma in conf/nutcracker.yml to use this feature.
I hope to hear your feedback about this patch.

### DIFF
--- a/conf/nutcracker.yml
+++ b/conf/nutcracker.yml
@@ -65,3 +65,18 @@ omega:
   servers:
    - 127.0.0.1:11214:100000
    - 127.0.0.1:11215:1
+
+sigma:
+  listen: 127.0.0.1:22125
+  hash: fnv1a_64
+  distribution: ketama
+  auto_eject_hosts: false
+  redis: true
+  server_retry_timeout: 2000
+  server_failure_limit: 1
+  servers:
+    - 127.0.0.1:6379:1 server1
+    - 127.0.0.1:6380:1 server2
+  sentinels:
+    - 127.0.0.1:26379:1
+    - 127.0.0.1:26380:1
--- a/src/nc_conf.c
+++ b/src/nc_conf.c
@@ -106,6 +106,10 @@ static struct command conf_commands[] = {
       conf_add_server,
       offsetof(struct conf_pool, server) },
 
+    { string("sentinels"),
+      conf_add_server,
+      offsetof(struct conf_pool, sentinel) },
+
     null_command
 };
 
@@ -213,6 +217,14 @@ conf_pool_init(struct conf_pool *cp, struct string *name)
                         sizeof(struct conf_server));
     if (status != NC_OK) {
         string_deinit(&cp->name);
+        return status;
+    }
+
+    status = array_init(&cp->sentinel, CONF_DEFAULT_SENTINELS,
+                        sizeof(struct conf_server));
+    if (status != NC_OK) {
+        string_deinit(&cp->name);
+        array_deinit(&cp->server);
         return status;
     }
 
@@ -352,6 +364,16 @@ conf_dump(struct conf *cf)
         for (j = 0; j < nserver; j++) {
             s = array_get(&cp->server, j);
             log_debug(LOG_VVERB, "    %.*s", s->len, s->data);
+        }
+
+        nserver = array_n(&cp->sentinel);
+        if (nserver > 0) {
+            log_debug(LOG_VVERB, "  sentinels: %"PRIu32"", nserver);
+
+            for (j = 0; j < nserver; j++) {
+                s = array_get(&cp->sentinel, j);
+                log_debug(LOG_VVERB, "    %.*s", s->len, s->data);
+            }
         }
     }
 }
@@ -964,8 +986,8 @@ conf_validate_structure(struct conf *cf)
 {
     rstatus_t status;
     int type, depth;
-    uint32_t i, count[CONF_MAX_DEPTH + 1];
-    bool done, error, seq;
+    uint32_t i, seq, count[CONF_MAX_DEPTH + 1];
+    bool done, error;
 
     status = conf_yaml_init(cf);
     if (status != NC_OK) {
@@ -974,7 +996,7 @@ conf_validate_structure(struct conf *cf)
 
     done = false;
     error = false;
-    seq = false;
+    seq = 0;
     depth = 0;
     for (i = 0; i < CONF_MAX_DEPTH + 1; i++) {
         count[i] = 0;
@@ -987,7 +1009,7 @@ conf_validate_structure(struct conf *cf)
      * keyx:
      *   key1: value1
      *   key2: value2
-     *   seq:
+     *   seq1:
      *     - elem1
      *     - elem2
      *     - elem3
@@ -996,11 +1018,15 @@ conf_validate_structure(struct conf *cf)
      * keyy:
      *   key1: value1
      *   key2: value2
-     *   seq:
+     *   seq1:
      *     - elem1
      *     - elem2
      *     - elem3
      *   key3: value3
+     *   seq2:
+     *     - elem1
+     *     - elem2
+     *     - elem3
      */
     do {
         status = conf_event_next(cf);
@@ -1039,8 +1065,8 @@ conf_validate_structure(struct conf *cf)
 
         case YAML_MAPPING_END_EVENT:
             if (depth == CONF_MAX_DEPTH) {
-                if (seq) {
-                    seq = false;
+                if (seq > 0) {
+                    seq = 0;
                 } else {
                     error = true;
                     log_error("conf: '%s' missing sequence directive at depth "
@@ -1052,10 +1078,10 @@ conf_validate_structure(struct conf *cf)
             break;
 
         case YAML_SEQUENCE_START_EVENT:
-            if (seq) {
+            if (seq >= CONF_MAX_SEQ) {
                 error = true;
-                log_error("conf: '%s' has more than one sequence directive",
-                          cf->fname);
+                log_error("conf: '%s' has more than %d sequence directive",
+                          cf->fname, CONF_MAX_SEQ);
             } else if (depth != CONF_MAX_DEPTH) {
                 error = true;
                 log_error("conf: '%s' has sequence at depth %d instead of %d",
@@ -1065,7 +1091,7 @@ conf_validate_structure(struct conf *cf)
                 log_error("conf: '%s' has invalid \"key:value\" at depth %d",
                           cf->fname, depth);
             }
-            seq = true;
+            seq++;
             break;
 
         case YAML_SEQUENCE_END_EVENT:
@@ -1192,6 +1218,50 @@ conf_validate_server(struct conf *cf, struct conf_pool *cp)
 }
 
 static rstatus_t
+conf_validate_sentinel(struct conf *cf, struct conf_pool *cp)
+{
+    uint32_t i, nsentinel;
+    bool valid;
+
+    nsentinel = array_n(&cp->sentinel);
+    if (nsentinel == 0)
+        return NC_OK;
+
+    if (!cp->redis) {
+        log_error("conf: pool '%.*s' is memcached, don't support sentinels",
+                  cp->name.len, cp->name.data);
+        return NC_ERROR;
+    }
+
+    /*
+     * Disallow duplicate sentinels - sentinels with identical "host:port:weight"
+     * or "name" combination are considered as duplicates. When server name
+     * is configured, we only check for duplicate "name" and not for duplicate
+     * "host:port:weight"
+     */
+    array_sort(&cp->sentinel, conf_server_name_cmp);
+    for (valid = true, i = 0; i < nsentinel - 1; i++) {
+        struct conf_server *cs1, *cs2;
+
+        cs1 = array_get(&cp->sentinel, i);
+        cs2 = array_get(&cp->sentinel, i + 1);
+
+        if (string_compare(&cs1->name, &cs2->name) == 0) {
+            log_error("conf: pool '%.*s' has sentinels with same name '%.*s'",
+                      cp->name.len, cp->name.data, cs1->name.len, 
+                      cs1->name.data);
+            valid = false;
+            break;
+        }
+    }
+    if (!valid) {
+        return NC_ERROR;
+    }
+
+    return NC_OK;
+}
+
+static rstatus_t
 conf_validate_pool(struct conf *cf, struct conf_pool *cp)
 {
     rstatus_t status;
@@ -1256,6 +1326,11 @@ conf_validate_pool(struct conf *cf, struct conf_pool *cp)
     }
 
     status = conf_validate_server(cf, cp);
+    if (status != NC_OK) {
+        return status;
+    }
+
+    status = conf_validate_sentinel(cf, cp);
     if (status != NC_OK) {
         return status;
     }
--- a/src/nc_conf.c
+++ b/src/nc_conf.c
@@ -274,6 +274,7 @@ conf_pool_each_transform(void *elem, void *data)
     TAILQ_INIT(&sp->c_conn_q);
 
     array_null(&sp->server);
+    array_null(&sp->sentinel);
     sp->ncontinuum = 0;
     sp->nserver_continuum = 0;
     sp->continuum = NULL;
@@ -309,6 +310,11 @@ conf_pool_each_transform(void *elem, void *data)
     sp->preconnect = cp->preconnect ? 1 : 0;
 
     status = server_init(&sp->server, &cp->server, sp);
+    if (status != NC_OK) {
+        return status;
+    }
+
+    status = server_init(&sp->sentinel, &cp->sentinel, sp);
     if (status != NC_OK) {
         return status;
     }
@@ -375,6 +381,144 @@ conf_dump(struct conf *cf)
                 log_debug(LOG_VVERB, "    %.*s", s->len, s->data);
             }
         }
+    }
+}
+
+static rstatus_t
+conf_write(FILE *fh, const char *fmt, ...)
+{
+    size_t size, n;
+    int len;
+    char buf[256];
+    va_list args;
+
+    len = 0;
+    size = sizeof(buf);
+    va_start(args, fmt);
+    len = nc_vscnprintf(buf, size, fmt, args);
+    va_end(args);
+
+    buf[len++] = '\n';
+
+    n = fwrite(buf, (size_t)len, (size_t)1, fh);
+    if (n == 0) {
+        return NC_ERROR;
+    }
+
+    return NC_OK;
+}
+
+void
+conf_rewrite(struct context *ctx)
+{
+    uint32_t i, j, npool, nserver;
+    struct conf *cf;
+    struct conf_pool *cp;
+    struct conf_server *cs;
+    struct string true_str, false_str, bool_str;
+    FILE *fh;
+    char tmp_conf_file[256];
+
+    cf = ctx->cf;
+
+    npool = array_n(&cf->pool);
+    if (npool == 0) {
+        return;
+    }
+
+    log_debug(LOG_VVERB, "%"PRIu32" pools in configuration file '%s'", npool,
+              cf->fname);
+
+    /* open tmp conf file to rewrite */
+    nc_snprintf(tmp_conf_file, 256, "%s.%d.tmp", cf->fname, (int) getpid());
+    fh = fopen(tmp_conf_file,"w");
+    if (fh == NULL) {
+        log_error("conf: failed to open tmp configuration file to rewrite");
+        return;
+    }
+
+    string_set_text(&true_str, "true");
+    string_set_text(&false_str, "false");
+
+    for (i = 0; i < npool; i++) {
+        cp = array_get(&cf->pool, i);
+
+        conf_write(fh, "%.*s:", cp->name.len, cp->name.data);
+        conf_write(fh, "  listen: %.*s",
+                  cp->listen.pname.len, cp->listen.pname.data);
+        conf_write(fh, "  hash: %.*s",
+                hash_strings[cp->hash].len, hash_strings[cp->hash].data);
+        if (cp->hash_tag.len > 0) {
+            conf_write(fh, "  hash_tag: \"%.*s\"", cp->hash_tag.len,
+                    cp->hash_tag.data);
+        }
+        conf_write(fh, "  distribution: %.*s",
+                dist_strings[cp->distribution].len, dist_strings[cp->distribution].data);
+
+        bool_str = cp->preconnect ? true_str : false_str;
+        conf_write(fh,"  preconnect: %.*s", bool_str.len, bool_str.data);
+
+        bool_str = cp->auto_eject_hosts ? true_str : false_str;
+        conf_write(fh, "  auto_eject_hosts: %.*s", bool_str.len, bool_str.data);
+
+        bool_str = cp->redis ? true_str : false_str;
+        conf_write(fh, "  redis: %.*s", bool_str.len, bool_str.data);
+        if (cp->timeout >= 0) {
+            conf_write(fh, "  timeout: %d", cp->timeout);
+        }
+        conf_write(fh, "  backlog: %d", cp->backlog);
+        conf_write(fh, "  client_connections: %d",
+                  cp->client_connections);
+        conf_write(fh, "  server_connections: %d",
+                  cp->server_connections);
+        conf_write(fh, "  server_retry_timeout: %d",
+                  cp->server_retry_timeout);
+        conf_write(fh, "  server_failure_limit: %d",
+                  cp->server_failure_limit);
+
+        nserver = array_n(&cp->server);
+        conf_write(fh, "  servers:");
+
+        for (j = 0; j < nserver; j++) {
+            cs = array_get(&cp->server, j);
+            if (cs->name.len >= cs->pname.len
+                    || nc_strncmp(cs->pname.data, cs->name.data, cs->name.len)) {
+                conf_write(fh, "   - %.*s %.*s",
+                        cs->pname.len, cs->pname.data,
+                        cs->name.len, cs->name.data);
+            } else {
+                conf_write(fh, "   - %.*s",
+                        cs->pname.len, cs->pname.data);
+            }
+        }
+
+        nserver = array_n(&cp->sentinel);
+        conf_write(fh, "  sentinels:");
+
+        for (j = 0; j < nserver; j++) {
+            cs = array_get(&cp->sentinel, j);
+            if (cs->name.len >= cs->pname.len
+                    || nc_strncmp(cs->pname.data, cs->name.data, cs->name.len)) {
+                conf_write(fh, "   - %.*s %.*s",
+                        cs->pname.len, cs->pname.data,
+                        cs->name.len, cs->name.data);
+            } else {
+                conf_write(fh, "   - %.*s",
+                        cs->pname.len, cs->pname.data);
+            }
+        }
+
+        /* write a empty line to conf when a pool is writen except the last one. */
+        if (i < npool - 1) {
+            conf_write(fh, "");
+        }
+    }
+
+    fclose(fh);
+
+    if (rename(tmp_conf_file, cf->fname) == -1) {
+        log_error("Error moving temp conf file on the final destination: %s", strerror(errno));
+        unlink(tmp_conf_file);
     }
 }
 
--- a/src/nc_conf.c
+++ b/src/nc_conf.c
@@ -284,6 +284,8 @@ conf_pool_each_transform(void *elem, void *data)
     sp->continuum = NULL;
     sp->nlive_server = 0;
     sp->next_rebuild = 0LL;
+    sp->next_sentinel_reconn = 0LL;
+    sp->sentinel_idx = 0;
 
     sp->name = cp->name;
     sp->addrstr = cp->listen.pname;
--- a/src/nc_conf.c
+++ b/src/nc_conf.c
@@ -168,6 +168,9 @@ conf_server_each_transform(void *elem, void *data)
     s->next_retry = 0LL;
     s->failure_count = 0;
 
+    /* proxy will set sentinel flag after server transform */
+    s->sentinel = 0;
+
     log_debug(LOG_VERB, "transform to server %"PRIu32" '%.*s'",
               s->idx, s->pname.len, s->pname.data);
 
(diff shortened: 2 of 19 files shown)


conf/nutcracker.yml
```diff
@@ -65,3 +65,18 @@ omega:
...
+  hash: fnv1a_64
+  distribution: ketama
+  auto_eject_hosts: false
+  redis: true
+  server_retry_timeout: 2000
+  server_failure_limit: 1
+  servers:
+    - 127.0.0.1:6379:1 server1
+    - 127.0.0.1:6380:1 server2
+  sentinels:
```
mishan: nit: don't you need three sentinels for quorum in this example?

andyqzb: @mishan You can configure any number of sentinels no matter how many sentinels you use. Becase twemproxy just need connect to one of the to fetch the redis address. 

Of cause, configuring all the sentinels you use is better. Because if some of the sentinels are dead, twemproxy can connect to the alive sentinels.

andyqzb: To avoid being confused, I added one sentinel address to the example configuration.

mishan:
This is looking really good. I'm going to see about testing this under load on Tagged and simulating some failovers to give it a real world test at scale.

andyqzb:
@mishan Great! I have tested the patch with some case, and use redis-benchmark to test it under load. But it's essential to test it under real world work load.

mishan:
@andyqzb Sorry for lack of updates from me, should get the chance to test this within a couple of weeks; made this a higher priority.

conf/nutcracker.yml
TheRealBill: There is actually a good reason to connect to multiple Sentinels. If the Sentinel you are connected to is on the minority side of a network partition, and the original master is in that partition, it could report it's "old" master. By querying all of the Sentinels you'd detect it and route to the majority master.

TheRealBill:
Question: If we're using sentinel. Why have Redis server entries in the config when we should be getting that from Sentinel?

andyqzb:
@TheRealBill It's a good idea, we can drop the servers config if we use the sentinel. While, we should think about some detail problems which I said in issue #297 . So, I don't realize it in the patch to keep it simple at the beginning. But it's a candidate feature in the future.

conf/nutcracker.yml
andyqzb: @TheRealBill In the partition, if twemproxy can only connect a part of the sentinels, and some told you new master, some told you old master. While twemproxy don't know whick is right if the both two sides don't have instances over half of all. Another, the design of sentinel doesn't work well in network partition, so I think take consideration of network partition is not needed.

TheRealBill: > On Mar 24, 2015, at 10:48 PM, andy <notifications@github.com> wrote:
> 
> In conf/nutcracker.yml <https://github.com/twitter/twemproxy/pull/324#discussion_r27093408>:
> 
> > @@ -65,3 +65,18 @@ omega:
> >    servers:
> >     - 127.0.0.1:11214:100000
> >     - 127.0.0.1:11215:1
> > +
> > +sigma:
> > +  listen: 127.0.0.1:22125
> > +  hash: fnv1a_64
> > +  distribution: ketama
> > +  auto_eject_hosts: false
> > +  redis: true
> > +  server_retry_timeout: 2000
> > +  server_failure_limit: 1
> > +  servers:
> > +    - 127.0.0.1:6379:1 server1
> > +    - 127.0.0.1:6380:1 server2
> > +  sentinels:
> @TheRealBill <https://github.com/TheRealBill> In the partition, if twemproxy can only connect a part of the sentinels, and some told you new master, some told you old master. While twemproxy don't know whick is right if the both two sides don't have instances over half of all. Another, the design of sentinel doesn't work well in network partition, so I think take consideration of network partition is not needed.
> 
> 

I think it is quite the contrary. If Twemproxy can’t verify the current master it should refuse to proxy commands for the server. If you’ve only got one sentinel you can communicate with, it would be wise to assume the current master is no longer the master, but minimally cautious to simply refuse the connections for that slot. That is preferable to having data go to multiple places. One of the *big* benefits to using Twemproxy+sentinel is the ability, if done correctly, to make split-brain scenario something the clients are immune to. I’ve done it outside of Twemproxy so it is certainly doable - and not terribly complicated.

Conceptually the best place to run those sentinels will be the Twemproxy nodes themselves. However that isn’t alway possible. As such your network between Sentinels and Sentinels+Redis can partition but twemproxy could still see the whole picture. But regardless by having the proxy *not* accept commands (or minimally not accept write commands) when you aren’t getting results which are confirmable you remove split-brain by nipping it at the very point it can happen - you stop proxying that node/slot.

The design of sentinel works fine in network partitions, it is the client which needs to be informed of changes, and Twemproxy is functionally the client in this scenario. Ideally, Twemproxy would subscribe to each sentinel’s master-change event channel.[1] Then, on a master-change event it would update that slot’s master setting and proxy to the new master. By only checking one sentinel you can’t reliably handle the case - and you don’t get the information as fast when you do get it. But by checking each of the sentinel servers you will either know you can’t communicate and can fail to deny-to-safe-mode, or catch the event from the majority’s side and know you need to reconfigure even though the other sentinel(s) hasn’t caught the change. Incidentally there is a lag between the failover and the non-leader Sentinels so even outside of a network split the proper route is to catch it from the leader.

Ideally you want these actions to be idempotent as well. In other words if you get a change notification multiple times, or the current pull, and the "new-master" is the same as last-updated-master, don’t make changes. I suspect you’ve already done that but sometime people miss that so I mention it solely out of that possibility. But *always* communicate with more than just one sentinel.

Perhaps to do this we could either:
1) specify a quorum of sentinels to communicate with
2) pull the list of known sentinels from sentinel and use that discovery to learn of and communicate with all of the other sentinels.

On the idea of not listing servers, perhaps we could do something like:
<podname> <servername> 
Where server name  is use by twemproxy for slot identification, but podname is the name you pass to sentinel to get the master & slaves (remember, a sentinel can manage multiple Redis master/slave combos). Then you can do *all* discovery via sentinel. 

On a related note I’m not seeing which code maps a given pod (“server” in Twemproxy parlance) to a given set of Sentinels. Each pod (server) may be managed by different sentinel pools. I’ve also not found where we allow you to use a pod-name other than the server name in the servers section. We should cover that as well. Perhaps something along the lines of (assuming we do not do full discovery via sentinel):
sentinel: true
servers:
   <servername> <podname> <list of sentinel:port pairs>
Or (better, IMO):
sentinel: true
servers:
   <servername> <podname1> <constellation1>
   <servername> <podname2> <constellation2>
sentiinels:
  IP:PORT:constellation1
  IP:PORT:constellation1
  IP:PORT:constellation1
  IP:PORT:constellation2
  IP:PORT:constellation2
  IP:PORT:constellation2

Where constellation = “the group of sentinels managing  a given pod”. This latter option could also work well for full-discovery in sentinel. I also think it looks the most clean and readable. We might consider quoting pod-name. IIRC you can pretty much use any ascii character other than a space in a name in sentinels.

Just throwing out some ideas to hopefully express better what I find as missing or ways we can do it even better. To say I use sentinel heavily might be a bit of an understatement. 

I see we currently have IP:port:weight for sentinels, but I hope we aren’t using weight at all - ideally we should not have it at all to avoid confusion.

It’d been a long day and I’ll have more time of the next week or two to really dig into this. I apologize for not getting to it sooner, and am really glad to see someone taking the first steps in making what I maintain to be a significant improvement to Twemproxy! So thanks for getting this ball rolling. :D

Cheers,
Bill


1. OK, *Ideally* I’d love to be able to fire twemproxy up unconfigured, then configure it entirely via API calls the same way you can w/Redis and Sentinel.

P.S. I’ll probably work on or find someone to work on adding Red Skull (http://redskull.io) connectivity as an alternative to Sentinel connectivity to simplify things - but not until I’ve added non-JSON API interfaces to Red Skull.

andyqzb: @TheRealBill  Thanks for your suggestions.
For the first suggestion, twemproxy should connect to all the sentinels. I think we shouldn't make twemproxy too complicated. It's just a client which supports sentinels. The problems in network partition should be solved by sentinel. For example, if the minority of sentinels don't server the clients, twemproxy will connect to a sentinel sunccessfully until it find a sentinel in majority part. Doing this in sentinel is simpler than doing it in twemproxy.
The podname suggestion is a good idea. I have thought to make mastername in sentinel as a mixed type like poolname-servername. So we can manage all the servers in different pool in the same sentinel. But the idea of specifying each server a different sentinel pool is not needed. I don’t think there are people want to config one pool's servers into different sentinel pools.
The weight in sentinel config is not used. I just want to reuse the code of loading server config in sentinel config load. Modifying the code to drop weight in sentinel load is better.

TheRealBill: > On Mar 25, 2015, at 12:14 PM, andy <notifications@github.com> wrote:
> 
> In conf/nutcracker.yml <https://github.com/twitter/twemproxy/pull/324#discussion_r27143045>:
> 
> > @@ -65,3 +65,18 @@ omega:
> >    servers:
> >     - 127.0.0.1:11214:100000
> >     - 127.0.0.1:11215:1
> > +
> > +sigma:
> > +  listen: 127.0.0.1:22125
> > +  hash: fnv1a_64
> > +  distribution: ketama
> > +  auto_eject_hosts: false
> > +  redis: true
> > +  server_retry_timeout: 2000
> > +  server_failure_limit: 1
> > +  servers:
> > +    - 127.0.0.1:6379:1 server1
> > +    - 127.0.0.1:6380:1 server2
> > +  sentinels:
> @TheRealBill <https://github.com/TheRealBill> Thanks for your suggestions.
> For the first suggestion, twemproxy should connect to all the sentinels. I think we shouldn't make twemproxy too complicated. It's just a client which supports sentinels.
> 
Not quite - it is a client of sentinel. Thus it should be following the guidelines at http://redis.io/topics/sentinel-clients

> The problems in network partition should be solved by sentinel. For example, if the minority of sentinels don't server the clients, twemproxy will connect to a sentinel sunccessfully until it find a sentinel in majority part.
> 
How does it determine if the sentinel it connects to is in a minority? In order for it to do that it would have to compare master settings across the sentinel constellation. You can’t just ask sentinel if it is in a minority network partition. You also can’t query the “known sentinels” info for the pod because that isn’t updated except on new sentinel discovery and pod resets.

If a net split does occur and the original master is in the minority partition, twemproxy will still be connected to it. When sentinel initiates a fail-over normally, it sends a client kill to the old master (if available) to DC existing connections. However, in this scenario that won’t DC twemproxy. Thus Twemproxy needs to be checking/monitoring for failovers and updating/reconnecting as appropriate. Anything short of that means to have reliable redundancy and avoid or minimize the split-brain scenario you have to use the current method of rewriting the config and restarting twemproxy.

It just occurred to me you might be trying to implement this as “just” a TCP proxy/load balancer to Sentinels. I surely hope that isn’t the case as you can’t do it reliably. The way the Sentinel and client setup works clients need direct access to every sentinel for the reasons I listed above regarding why Twemproxy would need to do the things I’m talking about. Clients need to connect to the sentinel and do more than just get the current master. They need PUBSUB and the ability to talk to each sentinel directly to ensure they aren’t talking to a minority sentinel which contained the original master. Please tell me you’re not trying to make Twemproxy a load balancer for Sentinel. :) It would be either as complicated or, more likely,  more complicated than what I am talking about above. If you’re using twemproxy you don’t want to talk directly to the backend redis instances - that defeats the main purpose of twemproxy.

> Doing this in sentinel is simpler than doing it in twemproxy.
> 
Except that Sentinel doesn’t control the clients directly. The mechanism for Sentinel helping clients is by either 1) having each client talk to and monitor every sentinel, or 2) having a script on the sentinels which reaches out to reconfigure the clients, or 3) by disconnecting clients connected after a reconfiguration of the pod (such as failovers). 

> The podname suggestion is a good idea. I have thought to make mastername in sentinel as a mixed type like poolname-servername. So we can manage all the servers in different pool in the same sentinel. But the idea of specifying each server a different sentinel pool is not needed. I don’t think there are people want to config one pool's servers into different sentinel pools.
> 
While it isn’t the pattern I generally recommend, especially in the case of twemproxy supporting sentinel, it is actually very common for people to have each pod run it’s own sentinels, and I consistently run into objections to doing it any other way. Furthermore in environments where a tool such as Red Skull is used to provide a cluster of managed sentinel constellations it would be the case there as well. Sentinel does start running into issues when you have high amounts of pods being monitored. So breaking that out across a banks of sentinel instances is also an expected use case. Not supporting it backs a potential user into a corner where you can increase the time it takes to failover. Given that Twemproxy is by design setup to proxy to essentially a bank of Redis instances there is a higher than normal likelihood of multiple sentinels being used.

You can learn some of these issues as https://github.com/antirez/redis/issues/2257 and https://github.com/antirez/redis/issues/2045 

> The weight in sentinel config is not used. I just want to reuse the code of loading server config in sentinel config load. Modifying the code to drop weight in sentinel load is better.
> 
Ok, sounds reasonable. You should update the docs for that section to be quite clear that weight isn’t used for sentinels. A few lines in the docs can be worth many questions as to why Twemproxy isn’t respecting the weights assigned. ;)

Cheers,
Bill

mishan:
@andyqzb I'm having some issue running this with redis-sentinel 2.9.11

[2015-03-25 16:04:36.736] nc.c:192 run, rabbit run / dig that hole, forget the sun / and when at last the work is done / don't sit down / it's time to dig another one
[2015-03-25 16:04:36.739] nc_sentinel.c:186 unknown server name:7
[2015-03-25 16:04:36.739] nc_sentinel.c:457 sentinel's response error, close sentinel conn.
[2015-03-25 16:04:36.739] nc_mbuf.c:300 mbuf read string failed, split char :
[2015-03-25 16:04:36.739] nc_sentinel.c:457 sentinel's response error, close sentinel conn.
[2015-03-25 16:04:36.739] nc_mbuf.c:300 mbuf read string failed, split char :
[2015-03-25 16:04:36.739] nc_sentinel.c:457 sentinel's response error, close sentinel conn.

Here's the config I'm using: http://pastebin.com/txtRVRj6

Any ideas?

andyqzb:
@mishan Maybe the output of sentinel has changed. I have tested the patch with sentinel 2.6 and 2.8. I will check this problem later.

andyqzb:
@mishan Maybe you config a server named 7, while you don't config it in the twemproxy. The twemproxy will pull all the server addresses in the sentinel and assume that all the server addresses got are configured in the twemproxy. So twemproy think there are errors if it gets some servers which are not configured in the twemproxy.

mishan:
@andyqzb Oh, my config just has the masters in it, not the slaves. That could explain it. I'll double check.

mishan:
Yeah, I see my problem now, actually. There was a mismatch between the naming convention in nutcracker.yml and the sentinel configs. Will fix and report back on results of testing. Thanks!

conf/nutcracker.yml
andyqzb: @TheRealBill I have seen the guidelines of sentinel clients. It said client just need to connect one of the sentinels just like what the patch does. Just like what the guidelines said below, It won't work well in the network partition.
>Note: it is possible that a stale master returns online at the same time a client contacts a stale Sentinel instance, so the client may connect with a stale master, and yet the ROLE output will match. However when the master is back again Sentinel will try to demote it to slave, triggering a new disconnection. The same reasoning applies to connecting to stale slaves that will get reconfigured to replicate with a different master.

I think twemproy shouldn't do too much things about distributed system. Sentinel is responsible for the cluster. So if you want the cluster works well under the network partition, the sentinels of minority should refuse to server the clients just like zookeeper or chubby.
I have seen the redis issue 2257. I think it's a suggestion to reduce the connections of sentinels. Sentinel is ok when it has 2000+ connections(under the issue condition). The epoll can process tens of thousands of connections easily. Maybe It's a problem when the cluster has thousands of redis.

TheRealBill: > On Mar 30, 2015, at 12:00, andy <notifications@github.com> wrote:
> 
> In conf/nutcracker.yml:
> 
> > @@ -65,3 +65,18 @@ omega:
> >    servers:
> >     - 127.0.0.1:11214:100000
> >     - 127.0.0.1:11215:1
> > +
> > +sigma:
> > +  listen: 127.0.0.1:22125
> > +  hash: fnv1a_64
> > +  distribution: ketama
> > +  auto_eject_hosts: false
> > +  redis: true
> > +  server_retry_timeout: 2000
> > +  server_failure_limit: 1
> > +  servers:
> > +    - 127.0.0.1:6379:1 server1
> > +    - 127.0.0.1:6380:1 server2
> > +  sentinels:
> @TheRealBill I have seen the guidelines of sentinel clients. It said client just need to connect one of the sentinels just like what the patch does. Just like what the guidelines said below, It won't work well in the network partition.
> 
But the client of twemproxy can't use sentinel directly because twemproxy is "in the way".  Since that was written we have developed operational knowledge and wisdom that tells us how to do it correctly, and it really isn't that hard. At a bare minimum you connect to each sentinel and pick the master reported by a majority of sentinels. If you can't reach said quorum, don't proxy. You can do it in a relatively simple script in HAProxy, so to argue it is overly complicated to do in Twemproxy doesn't hold IMO. It really is the least you should do. 
> Note: it is possible that a stale master returns online at the same time a client contacts a stale Sentinel instance, so the client may connect with a stale master, and yet the ROLE output will match. However when the master is back again Sentinel will try to demote it to slave, triggering a new disconnection. The same reasoning applies to connecting to stale slaves that will get reconfigured to replicate with a different master.
> 
> I think twemproy shouldn't do too much things about distributed system.
> 

Then don't use it with sentinel. By placing Twemproxy in front and having sentinel hidden from the clients you're preventing the clients from doing what you describe. You take on the responsibility of protecting the clients when you prevent them from protecting themselves. 

Doing the simple thing is not always the right thing. If you're going to make it impossible for the client to protect itself, you must take the precautions on it's behalf. Twemproxy partitions data, and with the addition of sentinel support it thus becomes a core part of a distributed system. At that point the assertion it shouldn't do the right thing with regard to proper behavior in distributed systems is rather moot. 
> Sentinel is responsible for the cluster. So if you want the cluster works well under the network partition, the sentinels of minority should refuse to server the clients just like zookeeper or chubby.
> 
Again, how exact do you think the clients can learn the sentinel they are talking to is in a minority partition? If you're going to propose they can, you have to show how. It also would be how sentinel would do it. There is one way: ask all sentinels you can connect to, and you can't get a majority of known sentinels, stop proxying the connections. 

You can wax on all day about what sentinel could or should do, but you're dealing with Twemproxy, not sentinel. I'm not sure you understand what sentinel is, as it doesn't serve clients it is a lookup service. 

When a network partition happens, sentinel has no way to know if a new master has been elected, and as such takes no action. It doesn't know if it is alone and all sentinels are alone (thus no master change happens) or it is in the only minority. And even then, it doesn't know who the new master is. Thus it can't know it is "in the minority" partition. Therefore it can't tell the clients to stop talking to the server as they may still be talking to the right server. 

But Twemproxy can figure it out. Any client can, until you put Twemproxy in front of it and prevent it from doing so. 
> I have seen the redis issue 2257. I think it's a suggestion to reduce the connections of sentinels. Sentinel is ok when it has 2000+ connections(under the issue condition). The epoll can process tens of thousands of connections easily. 
> 

The point of this issue, Andy, is showing that the condition you think is rare is actually common and in a non-trivial amount of the time required. This is just one of many conditions I've seen. I've seen thousands of these setups, multiple sets of sentinels across a bank of pods is not a rare condition. Nor is it going to become one anytime soon. 


Cheers,
Bill

andyqzb:
@mishan How about the testing?

mishan:
@andyqzb sorry haven't had the chance to try this with a production service yet, have had lots of work interrupts come up. I have it running on a development server now, probably will have time to test it next week on production. I'll let you know soon

idning:
Let's make this merged, @andyqzb  can you please add some cases with the test suite:

https://github.com/twitter/twemproxy/tree/master/tests

thekiki:
looking forward to this feature!

andyqzb:
@idning I'll add some sentinel test cases in the next week.

mishan:
I have this operationalized in a development environment, didn't want to rush it to production end of week. I want to run an experiment next week by putting this on production. It's more than just running your patch though, since there's been a bunch of nutcracker changes in between. But seems like it's a go to test on prod!

spyderdyne:
Def. Dont rush this 2 prod. Very fragile.

JamieCressey:
@mishan how did you get on testing this?

mishan:
@JamieCressey and everyone else, apologies I've made promises to have it on production earlier. I just got a house and moved and I've had things come up and naturally didn't want to rush it. I have it working *normally* on dev/stage.

I was just playing around with failing over on stage and am running into an issue with my application initially failing to connect to a shard, but the problem might either be that the .yml file doesn't get rewritten due to permissions (it's managed by Puppet, so we have to get around that) or that the application I'm using somehow doesn't handle it gracefully. Since twemproxy is a proxy, I don't see how the latter could be the case.

So, tl;dr: I need to do some more stage testing with manual failover as currently I haven't been entirely successful. Once that's green flag, I won't have hesitation pushing this to a prod tier that is not high risk but has plenty of volume

I have some limited time to poke at this tomorrow and then I'm out the next week. I really want to see this working, so I may end up spending some time next week or delegating the testing task to a teammate

More to come.

maralla:
Any news?

andyqzb:
@idning @mishan @maralla and everyone else, sorry for the delay of adding test cases. I just merged the updates from the master to my branch. @mishan can use the latest code to do the test. I will add the sentinel test cases in this weekend.

mishan:
@andyqzb I confirmed my issue was with not being able to write the new .yml file -- did some testing in a dev environment with sudden shutdown of all redis masters (worse cast scenario) and now it works. There's about 10 seconds after shutting down all the masters (20 of them) before the build I made off the latest of this branch before my requests stop returning errors talking to nutcracker.

    [2015-07-15 20:15:03.733] nc_response.c:118 s 29 active 0 is done
    [2015-07-15 20:15:03.757] nc_response.c:118 s 27 active 0 is done
    [2015-07-15 20:15:03.776] nc_response.c:118 s 35 active 0 is done
    [2015-07-15 20:15:03.800] nc_response.c:118 s 31 active 0 is done
    [2015-07-15 20:15:03.876] nc_response.c:118 s 33 active 0 is done
    [2015-07-15 20:15:14.958] nc_server.c:719 success switch sr-sdb000 to 10.98.20.105:6379:1
    [2015-07-15 20:15:15.366] nc_server.c:719 success switch sr-sdb018 to 10.98.20.105:6397:1
    [2015-07-15 20:15:15.980] nc_server.c:719 success switch sr-sdb003 to 10.98.20.105:6382:1
    [2015-07-15 20:15:16.059] nc_server.c:719 success switch sr-sdb002 to 10.98.20.105:6381:1
    [etc]

Granted it's on dev where everything is on small VMs instead of some high performance hardware. I'll see how it goes on stage tomorrow once I get the new build and init script ready and probably even end up putting on production for a work in progress tier. Even if there's 10 seconds of outage for a worst case scenario (all masters die), this is a huge improvement of what I have now!

Thanks!

andyqzb:
@mishan Do you config the sentienls's down-after-milliseconds to 10s? The proxy will switch its redis address to the new one immediately after the sentinels do the failover.

mishan:
@andyqzb Hrm, you called it! Not sure why that got end up set so high, 10 secs seems a little much, I'll have to look into this tomorrow.

Otherwise this is looking great then -- I've got a build ready to test on stage and hopefully production tomorrow; I'm feeling good about this!

Thanks

mishan:
Well it's happily running on production handling requests from an application to a Redis "cluster" (not Redis Cluster) doing 1MB/s -- tests on dev/stage I've done were pretty good. I haven't had a chance to do failover tests under production load yet, that's next.

andyqzb:
@idning I have added four test cases in the tests/test_system/test_sentinel.py. Maybe the test cases cost a little time when you run them, because they must wait the slave sync with master, the sentinel discover the slaves and process the failover.

mishan:
@andyqzb I've been using your branch nutcracker on prod for well over a month without any issues! This is great! I'd like to see this get merged to master.

A future enhancement I might suggest would be to add some new kind of runtime parameter to nutcracker that it can read its configs from read-only YML files and then maintain copies in, say, /var/run/nutcracker or something. I had to implement a hack to do this in the init.d script as our YML files are managed by puppet and I can't have nutcracker modify puppet-managed files.

But I would like to move that this gets merged sooner than later.

andyqzb:
@mishan I'm glad to hear your voice about the test in product environment.
@idning is reviewing the pr, let's wait for his response.

about the conf rewrite, I take example by the redis. maintaining multiple conf  will make user confused, so I don't think it's a good solution. Maybe you can use script triggerd by sentinel to modify the puppet files.

kri5:
Any updates?

BinWu1989:
@andyqzb , hi , is there any update about this feature? May I use the twemproxy as the loading balance proxy and redis sentinel as the solution of failover ?

andyqzb:
@BinWu1989 The patch is ready now. @mishan has used it on product environment. @idning told me he will review the patch, while he is too busy to do it these days.

BinWu1989:
@andyqzb the main purpose of this patch is to keep the master address in the twemproxy same with the sentinel , right ? Besides , do you have any idea about the loading balance to the slaves for the get requests ? I just want to use the sentinel for the solution for the failover and using the twemproxy as the read proxy , however, I was not pretty known about it , and is looking for advise.

aibibang:
Any updates?

therealbill:
@BinWu1989 for a read-only proxy just use HAProxy configured to talk to each node as a backend. Nothing hard or special/custom there.

mishan:
+1 for updates -- would love to see this get merged.

@aibibang You could also just check out this branch and build it. My company's running it on production without issues. One team has requested to start using it on their service to handle being able to write on master fail over.

aibibang:
@mishan thank you for your reply,When to release this feature ? I'm more concerned about this.
@andyqzb thanks

axot:
+1 a better solution instead using agent to update config file. When this will be merged into master?

andyqzb:
@axot @TrumanDu @mishan Thans for your attentions. @idning told me he is busy on his work these days, so the patch has to wait more time to be merged. @manjuraj , do you have time to review this patch?

jacan:
Any news, about this PR being merged?

kri5:
:+1:

mishan:
There's been no response from anyone at Twitter it seems. I forked it if(we)'s use and merged @andyqzb 's PR into our fork last year and we've been using it for a while now. I think I need to rebase it. I bumped the version to "0.5.0" https://github.com/ifwe/twemproxy if anyone's interested in giving it a try despite this PR not being merged.

stiller-leser:
Any updates on this? @mishan Are you still actively using @andyqzb patch in production?

mishan:
@stiller-leser We are still actively using it on production. One tier has been using it for the better part of a year now without issue. I'm now rolling it out to the rest of all our production tiers.

NuSkooler:
FWIW we are using @mishan's fork in development, and _very_ production soon (there are currently no known issues). This is greatly simplified over the old method of using additional agents to update configurations.

stiller-leser:
@mishan @NuSkooler Thanks for your feedback. Ultimately we decided to go with HAproxy to support mysql as well. Will keep an eye on Twemproxy though.

nasskach:
Why this feature was never merged ?! :(

src/nc_sentinel.c
```diff
@@ -0,0 +1,467 @@
...
+    string_init(&server_ip);
+    string_set_text(&sentinel_masters_prefix, "sentinel_masters");
+
+    line_buf = mbuf_get();
+    if (line_buf == NULL) {
+        goto error;
+    }
+
+    /* get sentinel master num at line 3 */
+    msg_read_line(msg, line_buf, 3);
```
TysonAndre: This will/does have issues if redis-sentinel adds new fields to the response, changes the order of the fields, or sends more lines than this expects. See https://github.com/twitter/twemproxy/issues/527 (wrong repo)

After upgrading from redis 2.x to 3.2.6 , I noticed this. I'll try to find a solution

FractalizeR:
Any news here on this PR? This looks promising because other solutions like this: https://github.com/yak0/twemsentinel look abandoned.

src/nc_sentinel.c
TysonAndre: https://github.com/ifwe/twemproxy/commits/build-nutredis contains a fix for this and other issues -  feel free to use https://github.com/ifwe/twemproxy/commit/9820caa202ab12bab01891174b6f66dbf37b9490

I've tested this with redis 3.2, and to some extent 4.0 (but not for long periods of time)

src/nc_conf.c
```diff
@@ -360,6 +384,158 @@ conf_dump(struct conf *cf)
...
+        return;
+    }
+
+    string_set_text(&true_str, "true");
+    string_set_text(&false_str, "false");
+
+    for (i = 0; i < npool; i++) {
+        cp = array_get(&cf->pool, i);
+
+        conf_write(fh, "%.*s:", cp->name.len, cp->name.data);
```
TysonAndre, line 460: This should check the error code of conf_write, in case the disk fills up or other errors occur, to avoid replacing configs with an empty file?

`#define conf_write_or_fail(...) if (conf_write(__VA_ARGS__) != NC_OK) { goto error; }`

See https://github.com/ifwe/twemproxy/pull/38 for an example of how that could be done

src/nc_sentinel.c
```diff
@@ -0,0 +1,471 @@
...
+        /* get server name */
+        status = mbuf_read_string(line_buf, ',', &server_name);
+        if (status != NC_OK) {
+            log_error("get server name failed.");
+            goto error;
+        }
+
+        server = server_find_by_name(ctx, sentinel->owner, &server_name);
+        if (server == NULL) {
+            log_error("unknown server name:%.*s", server_name.len, server_name.data);
```
TysonAndre, line 196: This should just be a continue and info/warning log. A sentinel server may refer to multiple redis servers, and the twemproxy config may only be a subset of them.

Developers might also temporarily set up additional servers with a given sentinel cluster to test.

Also, the sentinel patches were applied to the fork https://github.com/ifwe/twemproxy - which I'd planned to merge once I'd created a release for all of the other non-sentinel patches, though that's been delayed, and I expect resolving merge conflicts might take a while.

I should create a release - there were some edge cases I was trying to track down months ago, but I believe they weren't new edge cases caused by the latest patches

PR #326, opened 2015-03-04
issue 323: fix redis parse rsp bug for simple line 

charsyam:
in issue-323 redis return like this
*2
$2
ro
+OK

current twemproxy can't parse this type, so I append SW_SIMPLE for cover this.
Thank you.

### DIFF
--- a/src/proto/nc_redis.c
+++ b/src/proto/nc_redis.c
@@ -1687,6 +1687,7 @@ redis_parse_rsp(struct msg *r)
         SW_ERROR,
         SW_INTEGER,
         SW_INTEGER_START,
+        SW_SIMPLE,
         SW_BULK,
         SW_BULK_LF,
         SW_BULK_ARG,
@@ -1771,6 +1772,13 @@ redis_parse_rsp(struct msg *r)
             /* rsp_start <- p */
             state = SW_INTEGER_START;
             r->integer = 0;
+            break;
+
+        case SW_SIMPLE:
+            if (ch == CR) {
+              state = SW_MULTIBULK_ARGN_LF;
+              r->rnarg--;
+            }
             break;
 
         case SW_INTEGER_START:
@@ -1912,6 +1920,7 @@ redis_parse_rsp(struct msg *r)
                     /* response is '*0\r\n' */
                     goto done;
                 }
+
                 state = SW_MULTIBULK_ARGN_LEN;
                 break;
 
@@ -1953,9 +1962,17 @@ redis_parse_rsp(struct msg *r)
                     break;
                 }
 
-                if (ch != '$' && ch != ':') {
+                if (ch == ':' || ch == '+' || ch == '-') {
+                    /* handles not-found reply = '$-1' or integer reply = ':<num>' */
+                    /* and *2\r\n$2\r\nr0\r\n+OK\r\n or *1\r\n+OK\r\n */
+                    state = SW_SIMPLE;
+                    break;
+                }
+
+                if (ch != '$') {
                     goto error;
                 }
+
                 r->token = p;
                 r->rlen = 0;
             } else if (isdigit(ch)) {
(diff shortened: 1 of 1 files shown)


BrandonBrowning:
:shipit:

charsyam:
@manjuraj @idning Could you review this patch?

charsyam:
ping @manjuraj @idning this is bug patch. :)

idning:
@charsyam 

can you please give a link when redis return response like this?

I do not think this is valid.

charsyam:
@idning 
in redis rsp, one line response can exist. and also it in a part of response.
see this. http://redis.io/topics/protocol

part: RESP protocol description
```
In RESP, the type of some data depends on the first byte:
For Simple Strings the first byte of the reply is "+"
For Errors the first byte of the reply is "-"
For Integers the first byte of the reply is ":"
For Bulk Strings the first byte of the reply is "$"
For Arrays the first byte of the reply is "*"
```

and

Part: RESP Arrays

```
*5\r\n
:1\r\n
:2\r\n
:3\r\n
:4\r\n
$6\r\n
foobar\r\n
```

therealbill:
It would help if you listed the command to produce the response indicated. The response posted does not appear valid as '*' indicates an array, which can not be sent on a single line unless it is an empty array in which case the *entire* reply is "*0\r\n".

What is posted in the OP is not consistent with the protocol. Breaking it down:
Line 1: "*2\r\n" -> An array of two elements is coming
Line 2: "$2\r\n" -> first element: a string of length two is coming
Line 3: "ro\r\n" -> here is your string 'ro'
Line 4: "+OK"  -> ERROR

Returning "+OK" at line 4 is incorrect. The correct next item *must* be the next element in the 2 element array, not an "+OK" message. If Redis indeed sent that you need to submit the bug with reproduction details to the Redis repo, as that is not a valid response.

charsyam:
Hi, @therealbill 
I also agree with your opinion. It is error in semantic;
but I think it's grammar is valid.

so, we should parse this type.

in Redis protocol, there is a example like below:
```
*5\r\n
:1\r\n
:2\r\n
:3\r\n
:4\r\n
$6\r\n
foobar\r\n
```

therealbill:
@charsyam yes, in theory the following *should* parse:
`*5\r\n:1\r\n:2\r\n:3\r\n:4\r\n$6\r\nfoobar\r\n`

However, the proposed patch does not make this a parseable response.

I *have*  found where the original response listed is obtainable: In MULTI/EXEC. These commands are  still listed as not unsupported by Twemproxy. That is, so far, one of  only two places I've found where Redis will reply in that format (the other is to a subscriber when a message is published). If there is another command sequence which produces this response, please post it - it may or may not be a supported command and that should be taken into account.

The example response given is an array of arrays - which Redis has no actual support for as a datatype. However, knowing this is how transactions are handled it makes sense to have because Redis only replies with simple strings as a result of a command *which does not return anything else*. Returning an array with a simple string means the result of multiple commands in the transaction, and that what is actually returned is not a response but an array of responses - and each has to be parsed individually as responses, not as elements of an array or as a single response.

Note that pipelining does not create such a response structure - it merely sends everything back in a single flush as it would appear normally.

Thus in the example given the parser iterates over each as an individual response, hence the ":" identifying an integer *reply*, not a member of an actual list/set etc.. So as far as I can tell the only time Redis would respond in this fashion is when the server is sending the result of multiple commands in a single multi/exec transaction block - or to a subscriber when a message is published to the subscribed channel. PUBSUB is also listed as still unsupported.

BrandonBrowning:
I originally reported https://github.com/twitter/twemproxy/commit/9cf45ab02d29c6e9ce154a8950db743d1d263842#commitcomment-10031992 which should be fixed by this patch.  Does this help reproduce a real case without multi/exec or pubsub, or am I missing something?  I apologize as I can't completely understand the protocol discussions you're referring to @therealbill.

therealbill:
@BrandonBrowning yes, that fits perfectly as a valid, and I believe supported, case. Thanks.

idning:
Thanks @charsyam 

and I have add a test case for this issue: https://github.com/twitter/twemproxy/commit/786c6d4345e98d8f2bdb9a3e63cd666fc6f57615

PR #330, opened 2015-03-11
Add switch to use TCPKEEPALIVE option

charsyam:
I add tcpkeepalive option in pool.
to remove dead peer, I think it is better to add TCP KEEPALIVE
 
to use default value and system setting. I don't set other value.

### DIFF
--- a/src/nc_conf.c
+++ b/src/nc_conf.c
@@ -74,6 +74,10 @@ static struct command conf_commands[] = {
       conf_set_bool,
       offsetof(struct conf_pool, redis) },
 
+    { string("tcpkeepalive"),
+      conf_set_bool,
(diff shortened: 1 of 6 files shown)


idning:
great!

I think we can always enable KEEPALIVE (do not need a config switch to enable it)

https://github.com/twitter/twemproxy/pull/324 also has code ``nc_set_keepalive``: 

We may merge this PR after pull 324 merged.

manjuraj:
merged branch manually: https://github.com/twitter/twemproxy/commit/f449b5e9fa48c118167814a226d3e8e2868263d4

charsyam:
@manjuraj thanks.

PR #344, opened 2015-04-07
Support "touch" command for memcached and allow negative exptime

mckelvin:
The PR is similar to https://github.com/Netflix/dynomite/pull/76

* * *

The [touch command](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L318) is not supported in twemproxy. I'm not sure if there's any deliberate purpose, if so, please tell me.

Touch with negative exptime is supported by memcached while not well documented. I'm not familiar with its inner implements but I guess the proxy should better transport without any interruption.

```
 telnet localhost 11211
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
set foo 0 0 2
10
STORED
get foo
VALUE foo 0 2
10
END
touch foo -1
TOUCHED
get foo
END
```

### DIFF
--- a/notes/memcache.txt
+++ b/notes/memcache.txt
@@ -37,6 +37,10 @@
   where,
   <value> - uint64_t
 
+- Touch Command (touch):
+
+  touch <key> <expiry>[noreply]\r\n
+
 - Misc Commands (quit)
 
   quit\r\n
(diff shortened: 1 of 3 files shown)


manjuraj:
looks good. Could you update this file to include semantic for the touch command: https://github.com/twitter/twemproxy/blob/master/notes/memcache.txt as a part of this commit

mckelvin:
@manjuraj Updated.

PR #358, opened 2015-05-13
Remove the unix domain socket file before binding

xiaosuo:
We should remove the unix domain socket file before binding, otherwise
bind(2) will fail.

### DIFF
--- a/src/nc_stats.c
+++ b/src/nc_stats.c
@@ -834,6 +834,11 @@ stats_listen(struct stats *st)
         return NC_ERROR;
     }
 
+    if (si.family == AF_UNIX && si.addr.un.sun_path[0] == '/' &&
+        unlink(si.addr.un.sun_path) && errno != ENOENT) {
+        log_error("unlink %s failed: %s", si.addr.un.sun_path, strerror(errno));
+        return NC_ERROR;
+    }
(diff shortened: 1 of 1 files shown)


manjuraj:
this is not safe because the domain socket file could be in use by another process. Closing this for now; let me know if you disagree

xiaosuo:
AFAIK proxy_reuse() does the same thing. I think what you said shoud be avoided by the administrator.

xiaosuo:
AFAIK proxy_reuse() does the same thing. I think what you said shoud be avoided by the administrator.

xiaosuo:
AFAIK proxy_reuse() does the same thing. I think what you said shoud be avoided by the administrator.

xiaosuo:
AFAIK proxy_reuse() does the same thing. I think what you said shoud be avoided by the administrator.

manjuraj:
fair point; however stats are only exposed on a "ip:port" as of today and not on unix domain socket

xiaosuo:
In fact, twemproxy doesn't stop you from specifying a path for the -a option now, nonetheless the help message for -a is "set stats monitoring ip (default: 0.0.0.0)". I think there are people relying on this FEATURE besides me. Do we need to update the help message for -a?

PR #362, opened 2015-05-20
fix bug hang when command just has type only

charsyam:
In twemproxy.
command that just has type only(except ping/quit) cause hanging

ex)
```c
mset
```

so this patch return parse error with this command(not hang)

### DIFF
--- a/src/proto/nc_redis.c
+++ b/src/proto/nc_redis.c
@@ -406,6 +406,7 @@ redis_parse_req(struct msg *r)
                 if (r->rnarg == 0) {
                     goto error;
                 }
+
                 r->narg = r->rnarg;
                 r->narg_end = p;
                 r->token = NULL;
@@ -1068,6 +1069,8 @@ redis_parse_req(struct msg *r)
             case LF:
                 if (redis_argz(r)) {
                     goto done;
+                } else if (r->narg == 1) {
+                    goto error;
                 } else if (redis_argeval(r)) {
                     state = SW_ARG1_LEN;
                 } else {


charsyam:
ping @idning @manjuraj

charsyam:
@manjuraj @idning please review this

manjuraj:
I had to undo this commit because with this change the test "test_redis.test_protocol.test_wrong_argc" fails

charsyam:
@manjuraj Thank you for your effort.
but I think it is not patch's problem.
the test is bad.

in real redis, it returns err, 
```c
*1
$3
GET
-ERR wrong number of arguments for 'get' command
```

and previous memcached has bug that hang with command only :)
and the test just check timeout(for this bug)

```python
def test_wrong_argc():
    s = get_conn()

    s.sendall('*1\r\n$3\r\nGET\r\n')
    assert_fail('timed out', s.recv, 10000)
```

so we just fix this testcase.

I think just remove assert line and
uncommented below lines.

```python
    assert('' == s.recv(1000))  # peer is closed
    assert(data.startswith('-ERR wrong number of arguments'))
```

so this patch works fine. and test result expect connection close because of "bad argument"

src/proto/nc_redis.c
```diff
@@ -406,6 +406,7 @@ redis_parse_req(struct msg *r)
                 if (r->rnarg == 0) {
                     goto error;
                 }
+
```
manjuraj: delete this line

manjuraj:
great; could you submit a diff with the test case fixed. Not that I would be returning '-ERR Invalid argument'

charsyam:
@manjuraj some of my mistakes, I made new pr 
Please see https://github.com/twitter/twemproxy/pull/375

PR #363, opened 2015-05-21
fix bug crash when get command without key and whitespace

charsyam:
In twemproxy.
"get" command cause crash.

ex) after get there is space. '_' means space.
```c
get_  
```

so this patch return parse error when key len is 0 with get command


charsyam:
Ping @idning @manjuraj

manjuraj:
thx @charsyam

PR #367, opened 2015-06-11
fix crc16

MOON-CLJ:
nc_crc16 impl is not correct, eg, "FM|ivector_2129920" -> 2859837793, the result should be 43361.

crc16's return value should never beyond 65536, @manjuraj @idning, pls review.

refer to https://github.com/twitter/twemproxy/pull/313, and i dont know how do you think about this kind of error.

### DIFF
--- a/src/hashkit/nc_crc16.c
+++ b/src/hashkit/nc_crc16.c
@@ -56,7 +56,7 @@ uint32_t
 hash_crc16(const char *key, size_t key_length)
 {
     uint64_t x;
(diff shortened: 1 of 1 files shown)


mckelvin:
:+1:

Instead of modifing the `hash_crc16` code directly, I'd prefer to create a new `hash_crc16a` to  conform to https://github.com/twitter/twemproxy/commit/653b05f861ae3ce0ff143e1067af378f677a4d29 .

PR #378, opened 2015-06-23
Add REUSEPORT to proxy socket. 

wooparadog:
1. Allow twemproxy to work in multi-process, solving performance issue of using only 1 core.
2. Allow graceful reload in some way. (Start new process with new
   configuration, then kill the old one after some time. If twemproxy
   can stop listening on proxy sockets on specific signal, it'll work better.
   This mechanism is very similar to haproxy's)

### DIFF
--- a/src/nc_proxy.c
+++ b/src/nc_proxy.c
@@ -98,6 +98,7 @@ proxy_reuse(struct conn *p)
     switch (p->family) {
     case AF_INET:
     case AF_INET6:
+        nc_set_reuseport(p->sd);
         status = nc_set_reuseaddr(p->sd);
         break;
 
--- a/src/nc_util.c
+++ b/src/nc_util.c
@@ -64,6 +64,18 @@ nc_set_nonblocking(int sd)
 }
 
 int
+nc_set_reuseport(int sd)
+{
+    int reuse;
+    socklen_t len;
+
+    reuse = 1;
+    len = sizeof(reuse);
+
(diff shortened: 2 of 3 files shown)


wooparadog:
This feature is added in [Linux 3.9](https://lwn.net/Articles/542629/). So I should check if it actually exists, but I don't know the correct way to do it, again, my c sucks..

tevino:
Maybe something like this?

```diff
diff --git a/src/nc_core.h b/src/nc_core.h
index 3166ec9..edddc08 100644
--- a/src/nc_core.h
+++ b/src/nc_core.h
@@ -40,6 +40,12 @@
 # define NC_STATS 0
 #endif

+#ifdef HAVE_SO_REUSEPORT
+# define NC_SO_REUSEPORT 1
+#else
+# define NC_SO_REUSEPORT 0
+#endif
+
 #ifdef HAVE_EPOLL
 # define NC_HAVE_EPOLL 1
 #elif HAVE_KQUEUE
diff --git a/src/nc_util.c b/src/nc_util.c
index c6eca80..038af07 100644
--- a/src/nc_util.c
+++ b/src/nc_util.c
@@ -66,6 +66,8 @@ nc_set_nonblocking(int sd)
 int
 nc_set_reuseport(int sd)
 {
+    if (!NC_SO_REUSEPORT) return 0;
+
     int reuse;
     socklen_t len;
```

PR #387, opened 2015-07-08
Added Hootsuite as twemproxy users

anubhavmishra:
Hootsuite uses twemproxy in front of both Memcached and Redis services. Our blog post on our migration from MySQL and Memcached to Redis that uses Twemproxy is shown here: http://code.hootsuite.com/mysql-to-redis/


manjuraj:
Sweet. I read through the migration blog post. Congrats on a successful migration :) Seems like a non trivial effort.

My hunch regarding timeouts with logging "turned on" might have to do with logging to EBS volume on AWS boxes instead of a Instance store. AWS is notorious about throttling I/O to EBS.

PR #389, opened 2015-07-15
support redis bitpos command

charsyam:
redis support bitpos command from redis 2.8.7
we missed this command :)

http://redis.io/commands/bitpos


charsyam:
@manjuraj ping. :)

PR #393, opened 2015-07-22
support geo commands and nested multibulk reply.

charsyam:
1. support GEO Commands
   -> geoadd, geohash, geodist, georadius, georadiusbymember
2. support nested multibulk reply
   -> geo commands return nested multibulk reply

currently we only support this type of multibulk

```
                 * - mulit-bulk
                 *    - cursor
                 *    - mulit-bulk
                 *       - val1
                 *       - val2
                 *       - val3
```

but georadius and georadiusbymember returns maximum 3 depth multibulk. so this patch supports this kind of multibulk also.

```
                 * - mulit-bulk
                 *    - multi-bulk
                 *       - val1
                 *       - val2
                 *       - val3
                 *    - mulit-bulk
                 *       - val1
                 *       - val2
                 *       - val3
                 *
                 * - mulit-bulk
                 *    - multi-bulk
                 *       - val1
                 *       - multi-bulk
                 *          - val2
                 *          - val3
                 *    - mulit-bulk
                 *       - val1
                 *       - multi-bulk
                 *          - val2
                 *          - val3
```

PR #395, opened 2015-07-25
support nested multibulk reply

charsyam:
@manjuraj This is part of https://github.com/twitter/twemproxy/pull/393

first. I divided it that is only supporting nested multi bulk reply.

after merging this :)

I will PR second of them :)

Thanks for your reivew :)

### DIFF
--- a/src/nc_message.c
+++ b/src/nc_message.c
@@ -255,6 +255,7 @@ done:
     msg->narg_end = NULL;
     msg->narg = 0;
     msg->rnarg = 0;
+    msg->depth = 0;
     msg->rlen = 0;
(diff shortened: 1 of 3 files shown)


charsyam:
@manjuraj ping :)

charsyam:
@manjuraj I fixed bug for parsing nested multibunk reply.
and it passed tests :)

Thanks. :)

charsyam:
@manjuraj ping

TysonAndre:
Support for redis multibulks of arbitrary depth was added in #612 using a different approach - it looks like this PR is no longer necessary.

PR #396, opened 2015-08-04
Add SO_REUSEPORT support to socket

ton31337:
Add SO_REUSEPORT support to socket. Introduce new feature for Twemproxy running on newer kernels. With this feature you are able to do upgrades, restarts without any downtime. Kernel does load balancing between processes with the same host:port pairs.

### DIFF
--- a/src/nc_proxy.c
+++ b/src/nc_proxy.c
@@ -142,6 +142,16 @@ proxy_listen(struct context *ctx, struct conn *p)
         return NC_ERROR;
     }
 
+    if (pool->reuseport) {
+        status = nc_set_reuseport(p->sd);
+        if (status < 0) {
+            log_error("reuse of port '%.*s' for listening on p %d failed: %s",
+                      pool->addrstr.len, pool->addrstr.data, p->sd,
+                      strerror(errno));
+            return NC_ERROR;
+        }
+    }
+
     status = bind(p->sd, p->addr, p->addrlen);
     if (status < 0) {
         log_error("bind on p %d to addr '%.*s' failed: %s", p->sd,
@@ -294,7 +304,7 @@ proxy_accept(struct context *ctx, struct conn *p)
                 return NC_OK;
             }
 
-            /* 
+            /*
              * Workaround of https://github.com/twitter/twemproxy/issues/97
              *
              * We should never reach here because the check for conn_ncurr_cconn()
--- a/src/nc_stats.c
+++ b/src/nc_stats.c
@@ -827,7 +827,7 @@ stats_listen(struct stats *st)
         log_error("socket failed: %s", strerror(errno));
         return NC_ERROR;
     }
-
+    nc_set_reuseport(st->sd);
     status = nc_set_reuseaddr(st->sd);
     if (status < 0) {
         log_error("set reuseaddr on m %d failed: %s", st->sd, strerror(errno));
--- a/src/nc_util.c
+++ b/src/nc_util.c
@@ -75,6 +75,20 @@ nc_set_reuseaddr(int sd)
     return setsockopt(sd, SOL_SOCKET, SO_REUSEADDR, &reuse, len);
 }
 
+int
+nc_set_reuseport(int sd)
+{
+    int reuse;
+    socklen_t len;
+
+    reuse = 1;
+    len = sizeof(reuse);
+#ifdef SO_REUSEPORT
+    return setsockopt(sd, SOL_SOCKET, SO_REUSEPORT, &reuse, len);
+#endif
+    return -1;
+}
+
 /*
  * Disable Nagle algorithm on TCP socket.
  *
--- a/src/nc_conf.c
+++ b/src/nc_conf.c
@@ -78,6 +78,10 @@ static struct command conf_commands[] = {
       conf_set_bool,
       offsetof(struct conf_pool, tcpkeepalive) },
 
+    { string("reuseport"),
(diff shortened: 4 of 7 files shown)


ton31337:
@manjuraj didn't you have some time to take a look at this pr? :eyeglasses:

src/nc_stats.c
```diff
@@ -827,7 +828,9 @@ stats_listen(struct stats *st)
         log_error("socket failed: %s", strerror(errno));
         return NC_ERROR;
     }
-
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3,9,0)
```
charsyam: @ton31337 twemproxy can run on mac or freebsd also. so I think you should check other os too not only linux :)

ton31337: @charsyam :+1:

ton31337:
@manjuraj why it's not possible to check why checks have failed?

src/nc_stats.c
ton31337: @charsyam are you able to test this on FreeBSD ? :)

ton31337:
tagging @manjuraj

ton31337:
@manjuraj anything is missing here? some input from you would be useful..

manjuraj:
change looks great - i need to test this before merging.

ton31337:
@manjuraj how can I help you to test this?

ton31337:
@manjuraj looks like CentOS, RHEL had already backported SO_REUSEPORT option even in 2.6.32 kernels. This makes more sense.

ton31337:
ping @manjuraj

ofirule:
+1
we also implemented the same feature: https://github.com/twitter/twemproxy/pull/435
Just noticed your pull request

ton31337:
yeah, it's even back ported to CentOS versions of 2.6 kernels..

ton31337:
@manjuraj ping

src/nc_util.c
```diff
@@ -75,6 +78,24 @@ nc_set_reuseaddr(int sd)
...
 }
 
+int
+nc_set_reuseport(int sd)
+{
+    int reuse;
+    socklen_t len;
+
+    reuse = 1;
+    len = sizeof(reuse);
```
TysonAndre, line 88: I think this should be replaced with `#ifdef SO_REUSEPORT`, which would cover all of the OS cases

src/nc_util.c
```diff
@@ -22,6 +22,9 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <netdb.h>
+#ifdef __linux
+#include <linux/version.h>
```
TysonAndre, line 26: The `__linux` specific code can be removed if `#ifdef SO_REUSEPORT` is used instead.

src/nc_proxy.c
```diff
@@ -134,6 +134,9 @@ proxy_listen(struct context *ctx, struct conn *p)
         return NC_ERROR;
     }
 
+    if (pool->reuseport)
+        nc_set_reuseport(p->sd);
```
TysonAndre, line 138: 1. This conditional can be moved into `proxy_reuse` - if reuseport config is set and this fails, then NC_ERROR should be returned
2. This only seems to have a documented effect for AF_INET/AF_INET6 https://man7.org/linux/man-pages/man7/socket.7.html

```
    switch (p->family) {
    case AF_INET:
    case AF_INET6:
        status = nc_set_reuseaddr(p->sd);
        break;
```

src/nc_stats.c
```diff
@@ -827,7 +827,7 @@ stats_listen(struct stats *st)
         log_error("socket failed: %s", strerror(errno));
         return NC_ERROR;
     }
-
+    nc_set_reuseport(st->sd);
```
TysonAndre, line 830: Looks good, SO_REUSEPORT would definitely help with errors I've encountered during startup when stats port is listening on all interfaces.

TysonAndre (changes requested):

src/nc_util.c
ton31337: Thanks, fixed, it was quite an old commit :)

src/nc_proxy.c
ton31337: I moved to be the same logic as for some other socket options, like `nc_set_nonblocking` or so.

src/nc_util.c
```diff
@@ -22,6 +22,9 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <netdb.h>
+#ifdef SO_REUSEPORT
+#include <linux/version.h>
+#endif
```
TysonAndre, line 27:
```suggestion
```

I meant remove the change to includes entirely.

This will also fail to build on non-linux platforms, which is why there was originally the `#ifdef __linux` check in a previous revision of your PR

TysonAndre (changes requested):

src/nc_util.c
ton31337: Roger, fixed.

TysonAndre:
> change looks great - i need to test this before merging.

This seems to work locally - testing since I haven't used SO_REUSEPORT elsewhere. I can start two nutcracker instances for memcached with `reuseport: true` and
- stats ports are shared and go to one of the two instances
- New connections are shared and go to one of the two instances
- Connections stay connected to the same instance

ton31337:
@TysonAndre do I need to do something else here?

PR #397, opened 2015-08-05
Document the client_connections config option

tom-dalton-fanduel:
We're using this setting, so thought it should be documented.

I had a very quick look at the code and couldn't immediately see where there's a check that the number of connections isn't exceeded, but I assume it's in there somewhere!


orimarti:
#477

PR #411, opened 2015-09-08
add rbtree_entry to fetch a struct pointer from a rbtree node pointer

ideal:
So the code is more generic.


ideal:
屏幕快照 2019-07-18 下午11 56 35

Seems that there is problem on signing page?

PR #412, opened 2015-09-09
Removing duplicate Hootsuite mention in list of companies using Twemproxy

dec5e:
Hootsuite was added twice: in #291 and #387.

PS. Maybe it's good to sort companies list alphabetically to prevent such duplication in future. I can do it if you agree.


dec5e:
ping @manjuraj

dec5e:
The project seems abandoned

PR #406, opened 2015-08-24
Fix parsing bug when error body contains no spaces

tom-dalton-fanduel:
Fix for issue https://github.com/twitter/twemproxy/issues/404

Parsing a redis error messages with no space in the body results in [part of] the following message being consumed and resulting errors.

PR #423, opened 2015-10-20
Fix parsing bug when error body contains no spaces

charsyam:
This is only copy for https://github.com/twitter/twemproxy/pull/406 (by @tom-dalton-fanduel )
for testing. I just PR again.

### DIFF
--- a/src/proto/nc_redis.c
+++ b/src/proto/nc_redis.c
@@ -1920,6 +1920,11 @@ redis_parse_rsp(struct msg *r)
 
                     break;
                 }
+
+                if (ch == CR) {
+                    p -= 1;
+                }
+
                 state = SW_RUNTO_CRLF;
             }
 
--- a/tests/test_redis/test_lua_error_return.py
+++ b/tests/test_redis/test_lua_error_return.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python
+#coding: utf-8
+
+from unittest import TestCase
+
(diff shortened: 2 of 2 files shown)


charsyam:
@manjuraj This passed testcases. How about recommiting it?

charsyam:
@manjuraj ping

tom-dalton-fanduel:
+1 from me too!

charsyam:
@manjuraj ping again :)

src/proto/nc_redis.c
```diff
@@ -1920,6 +1920,11 @@ redis_parse_rsp(struct msg *r)
 
                     break;
                 }
+
+                if (ch == CR) {
```
TysonAndre, line 1924: An alternative from 3e55462c0e5d9d95a88e099a777b7b3ebd373065 in a fork

```
-                state = SW_RUNTO_CRLF;
+                if (ch == '\r') {
+                    state = SW_ALMOST_DONE;
+                } else {
+                    // Read remaining characters until '\r'
+                    state = SW_RUNTO_CRLF;
+                }
             }
```

Both approaches work.

TysonAndre:
The fix was moved to https://github.com/twitter/twemproxy/pull/611 along with a C unit test and updating it to support the newer python/library versions for the integration test, avoiding going backwards in a state machine out of personal preference.

I'd found the same bug before seeing this PR and fixed it in a slightly different way, though both work

PR #431, opened 2015-11-12
Allow ignoring SELECT from client

rosmo:
Even if you are just using database 0, some developers want to call Redis' SELECT command always up front. This add a new redis_ignore_select option to pools to simply return OK for any SELECT command.

### DIFF
--- a/src/proto/nc_redis.c
+++ b/src/proto/nc_redis.c
@@ -62,6 +62,8 @@ static bool
 redis_arg0(struct msg *r)
 {
     switch (r->type) {
+    case MSG_REQ_REDIS_SELECT:
+
     case MSG_REQ_REDIS_EXISTS:
     case MSG_REQ_REDIS_PERSIST:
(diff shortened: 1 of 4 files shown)


Fgruntjes:
+1 would really like to see this merged, this prevents us from using twemproxy i.c.m. with Magento sessions.

src/proto/nc_redis.c
```diff
@@ -2695,7 +2704,11 @@ redis_reply(struct msg *r)
     switch (r->type) {
     case MSG_REQ_REDIS_PING:
         return msg_append(response, rsp_pong.data, rsp_pong.len);
-
+    case MSG_REQ_REDIS_SELECT:
```
TysonAndre, line 2707: Alternately or additionally, it may be safe to allow forwarding this entirely if the db number is equivalent to the one in the config file for redis_db (`pool->redis_db`)

PR #435, opened 2015-11-23
Added support for listening to same port from multiple twemproxy processes

ofirule:
My team (from IBM Security Trusteer) has ran into some bottlenecks when using twemproxy and developed this patch.

This patch allows to use socket option SO_REUSEPORT on twemproxy socket. This option is available on Linux kernel version 3.9+.

After stress testing our redis-server we found that handling many clients consumes too much processing from redis but not redis code itself rather its networking code (epoll/etc.).

It seems that the single thread nature of redis maxes out the networking part with one core in extreme conditions (like many connections scenario, no pipelining, etc.).

Using this patch we can connect many twemproxy instances (which runs on the same computer) to a redis server using the same socket, we were able to get a much greater throughput and still have one logical redis server. 

These twemproxy instances can all be configured with the same configuration file, and only needs to have a different stats port. (we don't enable SO_REUSEPORT for that socket)


muayyad-alsadi:
+1

merge please

Fleurer:
+1

PR #461, opened 2016-03-22
Fix memory leak for redis mset.

huachaohuang:


andyqzb:
I merge the bugfix in #486 , but using mbuf_length on a mbuf is not a good way after putting the mbuf to the free list. So can you create a new pr based on the master, I want to merge your code.

.travis.yml
```diff
@@ -1,3 +1,3 @@
 language: c
 script: bash ./travis.sh
-
+sudo: required
```
andyqzb: What is this line doing?

TysonAndre:
> I merge the bugfix in 486 , but using mbuf_length on a mbuf is not a good way after putting the mbuf to the free list. So can you create a new pr based on the master, I want to merge your code.

My understanding of this: twemproxy is single threaded so it doesn't cause issues in practice, but would agree in principle (e.g. if mbuf_free were to be replaced with an actuall free call as a debug compile option to debug memory leaks or use-after free)

PR #462, opened 2016-03-22
Fix a bug for msg->mlen

deep011:
Function msg_append() already added the dst->mlen.


TysonAndre:
This pr seems to be correct, function msg_append() already added the dst->mlen.

A message (dst) is made up of 1 or more mbufs, and the `bytes` is the length of the msg, which is the sum of the length of the mbufs. msg_append/mbuf_insert add individual mbufs from the source to the destination.

As for how twemproxy is currently working despite that, I expect it's because mlen isn't used and that the code that generates redis responses acts on the mbuf list - the length might only be used for logging


> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our Contributor License Agreement before we can accept your contribution.

PR #480, opened 2016-07-12
Implement ERROR protocol

willfitch:
Rather than severing connections upon an invalid event, this patch adds the ERROR portion of the Memcached protocol.  This does not, however, add CLIENT_ERROR or SERVER_ERROR.  That can be discussed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@
 *.~
 *.project
 *.cproject
+.settings/
 
 # Core and executable
 core*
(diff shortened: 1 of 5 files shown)


willfitch:
Will update tests to check for error rather than closed connection

TysonAndre:
1. If the client sends a malformed command, it's possible the responses and requests will be out of sync, so closing the connection immediately makes the most sense. Not closing connections on invalid requests will introduce new failure modes.
2. A better solution may be upgrading the memcached library/application to avoid sending requests with invalid memcached keys - this may have been done in the last 5 years
3. I'm an external collaborator, but twitter requires CLAs for PRs starting in 2019

PR #486, opened 2016-08-03
fix a memory leak bug for mset command

deep011:


andyqzb:
@deep011 good job!

qunchenmy:
Great！

PR #496, opened 2016-09-20
Docker setup

BlackVegetable:
Our fork off of twemproxy is simply to allow
us to use our own configuration files within
our own image in Docker.

It is times like these that I hate not rebasing at Keen. ;)


BlackVegetable:
It would help if I looked at which repository I was merging into. Doh!

PR #497, opened 2016-09-26
Added Geo Functionality

mahdi-hdi:
New Redis functionality for geo date type now supported:

```
GEOADD
GEODIST
GEOHASH
GEOPOS
GEORADIUS
GEORADIUSBYMEMBER
```

**Mutli bulk array** was implemented in Redis response parser to support response of **GEORADIUS**.

### DIFF
--- a/src/nc_message.c
+++ b/src/nc_message.c
@@ -255,6 +255,7 @@ done:
     msg->narg_end = NULL;
     msg->narg = 0;
     msg->rnarg = 0;
+    SLIST_INIT(&msg->argstk_head);
     msg->rlen = 0;
     msg->integer = 0;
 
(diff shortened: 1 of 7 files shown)


manjuraj:
@andyqzb ^

andyqzb:
@mahdi-hdi Can you fix your test cases of geo cmd?

mahdi-hdi:
@andyqzb The test cases use redis-py but the problem is that they didn't release GEO functionality yet (lastest version is 2.10.5). So, you must fetch their master branch in case of testing which I prefer don't add it to travis.sh.

tlhunter:
Bump

TysonAndre:
Support for parsing geo commands was added as part of a larger change to add support for many new/updated redis commands based on the latest commands.json published for redis, see 1239ef9e2f14f954b0f5da22369c0d994a347e45 and https://github.com/twitter/twemproxy/pull/599 (etc.)

A more generic approach was also added yesterday to support parsing multi-bulk responses of arbitrary depths and shapes

PR #565, opened 2018-09-30
Summary: this PR intends to make twemporxy able to parse nested multibulks

qingping209:
i.e. the response of georadiuswithcoord. 

Problem

 twemporxy is not able to parse nested multibulks reponse, which is not necessary before geo introduced but now we need it.  Instead of using an ad hoc hack to parse the response of geo  command, i.e. georadiuswithcoord, it would be better to have a generic approach support any syntactically correct redis response of multibulks, which can make  twemproxy work with new redis commands at no cost in the future. 

Solution

This is achieved by introducing a small stack in struct msg, this stack remembers the number of arguments to be parsed of current multibulk and all nesting multibulks.   Changes only affect the parse path of multinulk response, and everything else is untouched. 

to test geodist between nonexisting members of a zset, checkout this PR: https://github.com/andymccurdy/redis-py/pull/1035 to make the test work.

Result

This PR make twemproxy work with all geo commands.  I have a in house redis-server which can produce  multibulk responses of various combinations, and my twemproxy instance with this patch is happy with them.  here are the  response combinations(with the max depth of 3, and testmultibulk 6  with a depth of 4)

127.0.0.1:55555> testmultibulk 0
(empty list or set)

127.0.0.1:55555> testmultibulk 1
1) (nil)
2) (integer) 666

127.0.0.1:55555> testmultibulk 2
1) (empty list or set)

127.0.0.1:55555> testmultibulk 3
1) (nil)
2) (integer) 666
3) 1) (nil)
   2) "hello"

127.0.0.1:55555> testmultibulk 4
1) (nil)
2) (nil)
3) (integer) 666
4) 1) 1) "biubiubiu"
   2) 1) (nil)
   3) 1) OK

127.0.0.1:55555> testmultibulk 5
1) (nil)
2) (nil)
3) (integer) 666
4) 1) 1) "biubiubiu"
   2) 1) (nil)
   3) OK
5) 1) (integer) 666
   2) (error) OOM command not allowed when used memory > 'maxmemory'.
   3) OK

127.0.0.1:55555> testmultibulk 6 
(error) ERR Invalid argument

127.0.0.1:55555> testmultibulk 7
1) (nil)
2) (nil)
3) (integer) 666
4) 1) 1) "biubiubiu"
   2) 1) (nil)
   3) OK
5) 1) (integer) 666
   2) (error) OOM command not allowed when used memory > 'maxmemory'.
   3) OK
6) 1) 1) OK
7) 1) (empty list or set)
   2) (empty list or set)
   3) (empty list or set)

### DIFF
--- a/src/nc_message.c
+++ b/src/nc_message.c
@@ -255,6 +255,8 @@ done:
     msg->narg_end = NULL;
     msg->narg = 0;
     msg->rnarg = 0;
+    memset(msg->stack, 0, sizeof(msg->stack));
+    msg->nested_depth = 0;
     msg->rlen = 0;
     msg->integer = 0;
 
--- a/src/nc_message.h
+++ b/src/nc_message.h
@@ -20,6 +20,8 @@
 
 #include <nc_core.h>
 
+#define MAXDEPTH 3
+
 typedef void (*msg_parse_t)(struct msg *);
 typedef rstatus_t (*msg_add_auth_t)(struct context *ctx, struct conn *c_conn, struct conn *s_conn);
 typedef rstatus_t (*msg_fragment_t)(struct msg *, uint32_t, struct msg_tqh *);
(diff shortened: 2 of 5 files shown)


charsyam:
@qingping209 I think it is duplicated with https://github.com/twitter/twemproxy/pull/395

qingping209:
> 
> 
> @qingping209 I think it is duplicated with #395

oops ...   we pretty much did the same thing.  Is this project still in mainenace ?

TysonAndre:
This commit was merged as part of https://github.com/twitter/twemproxy/pull/612 which also removes the limitation on stack depth

I hadn't gotten around to adding a standardized way to skip integration tests if the redis server version was too old - https://github.com/twitter/twemproxy/blob/master/.github/workflows/main.yml also tests redis 3.0 and georadius was added in 3.2 - https://redis.io/commands/georadius

(multiple versions of redis server are tested against to prepare for redis sentinel support - redis sentinel's response format differs slightly in different redis versions)

PR #567, opened 2018-11-22
FIX: ketama isn't stable when redis/mc addr length is longer than KETAMA_MAX_HOSTLEN

liguangbo:
## Problem:

Ketama result is not stable. Distribute the same key to different Redis/Memcached server when twemproxy restart.

## Reason:

When we use domain to describe a redis server in twemproxy, server name maybe too long and exceed the KETAMA_MAX_HOSTLEN=86. 

host is consist of redis/memcached addr, port and continuum pointer_index.
```
hostlen = snprintf(host, KETAMA_MAX_HOSTLEN, "%.*s-%u",
                               server->name.len, server->name.data,
                               pointer_index - 1);
```

An addr may be a ip addr or domain name. According to the [RFC1035](https://tools.ietf.org/html/rfc1035), domain name length is 255 or less. So the KETAMA_MAX_HOSTLEN should be 
```
255(domain length)+1(:)+5(port max length)+1(-)+10(pointer_index max length)+1(\0)=273
```

The return value of snprintf maybe exceed KETAMA_MAX_HOSTLEN. 

>The snprintf() and vsnprintf() functions will write at most size-1 of the characters printed into the output string (the size'th character then gets the terminating `\0'); if the return value is greater than or equal to the size argument, the string was too short and some of the printed characters were dis-carded.  The output is always null-terminated, unless size is 0.

Use the return value of snprintf to do md5_signature could memcpy random data. So the ketama result is not stable.

```
void
MD5_Update(MD5_CTX *ctx, void *data, unsigned long size)
{
    MD5_u32plus saved_lo;
    unsigned long used, free;

    saved_lo = ctx->lo;
    if ((ctx->lo = (saved_lo + size) & 0x1fffffff) < saved_lo) {
        ctx->hi++;
    }
    ctx->hi += size >> 29;

    used = saved_lo & 0x3f;

    if (used) {
        free = 64 - used;

        if (size < free) {
            memcpy(&ctx->buffer[used], data, size);
            return;
        }

        memcpy(&ctx->buffer[used], data, free);
        data = (unsigned char *)data + free;
        size -= free;
        body(ctx, ctx->buffer, 64);
    }

    if (size >= 64) {
        data = body(ctx, data, size & ~(unsigned long)0x3f);
        size &= 0x3f;
    }

    memcpy(ctx->buffer, data, size);
}

```

## Solution:

To fix this, I change KETAMA_MAX_HOSTLEN to 273 and check if hostlen > KETAMA_MAX_HOSTLEN.

### DIFF
--- a/src/hashkit/nc_ketama.c
+++ b/src/hashkit/nc_ketama.c
@@ -25,7 +25,7 @@
 
 #define KETAMA_CONTINUUM_ADDITION   10  /* # extra slots to build into continuum */
 #define KETAMA_POINTS_PER_SERVER    160 /* 40 points per hash */
-#define KETAMA_MAX_HOSTLEN          86
+#define KETAMA_MAX_HOSTLEN          273 /* 273 is 255(domain or ip)+1(:)+5(port)+1(-)+10(uint32)+1(\0) */
 
 static uint32_t
 ketama_hash(const char *key, size_t key_length, uint32_t alignment)
@@ -178,6 +178,9 @@ ketama_update(struct server_pool *pool)
             hostlen = snprintf(host, KETAMA_MAX_HOSTLEN, "%.*s-%u",
                                server->name.len, server->name.data,
                                pointer_index - 1);
+            if (hostlen > KETAMA_MAX_HOSTLEN) {
+                hostlen = KETAMA_MAX_HOSTLEN;
+            }
 
             for (x = 0; x < pointer_per_hash; x++) {
                 value = ketama_hash(host, hostlen, x);


git-hulk:
@manjuraj anyone can review this pr?

axot:
We also hit the issue, does increase max length is safe?

git-hulk:
@axot yeah, we fixed it  in [meitu/twemproxy](https://github.com/meitu/twemproxy) and working fine in production

axot:
Hi @git-hulk it is a good news to hear it will work, I’m interesting does meitu use meitu/twemproxy in production? I’m also consider to use it with hot config reload and many fixes.

git-hulk:
@axot  yeah, also some others using `meitu/twemproxy` on production. we supported the multi-process and hot config reload, as well as request/server latency to stats on `twitter/twemproxy`, it works fine for us. I'm very glad that you would try it. :)

axot:
@git-hulk It sounds great, I will try your version in our production.

TysonAndre:
LGTM - I'm not a maintainer but I'm somewhat familiar with the code - this shouldn't have any performance impact because server_update is only called when a server is added or removed from the pool, which should be infrequent

src/hashkit/nc_ketama.c
```diff
@@ -178,6 +178,9 @@ ketama_update(struct server_pool *pool)
             hostlen = snprintf(host, KETAMA_MAX_HOSTLEN, "%.*s-%u",
                                server->name.len, server->name.data,
                                pointer_index - 1);
+            if (hostlen > KETAMA_MAX_HOSTLEN) {
+                hostlen = KETAMA_MAX_HOSTLEN;
```
TysonAndre, line 182: nit: I strongly doubt a use case would exceed the new host length in practice, but if it did, the last byte at `host[KETAMA_MAX_HOSTLEN - 1]` would be `'\0'`, and the implementation as-is would be deterministic it may differ from other implementations for hosts of length much longer than 255

> If the resulting string would be longer than **n-1 characters,** the remaining characters are discarded and not stored, but counted for the value returned by the function.

I can't see that mattering unless OS dns name limits are ever raised from https://www.freesoft.org/CIE/RFC/1035/9.htm

```suggestion
            if (hostlen >= KETAMA_MAX_HOSTLEN) {
                hostlen = KETAMA_MAX_HOSTLEN - 1;
```

TysonAndre:
Closed in https://github.com/twitter/twemproxy/commit/cb81d61ae3fa03a634fe331c546995ec7e3c0464 with the amendment suggested in https://github.com/twitter/twemproxy/pull/567#discussion_r611866867

PR #569, opened 2018-12-21
Calculate md5 signature only once per point

yatsukhnenko:
This PR contains optimisation of creating the ring for ketama algorithm. MD5 hash calculates only once per point instead of calculation in all iterations.


yatsukhnenko:
@manjuraj, @idning

yatsukhnenko:
Up @idning @manjuraj

PR #587, opened 2020-05-25
support multi-bulk-result

kukey:
support multi-bulk-result

### DIFF
--- a/src/nc_message.c
+++ b/src/nc_message.c
@@ -256,6 +256,7 @@ done:
     msg->narg = 0;
     msg->rnarg = 0;
     msg->rlen = 0;
+    msg->curr_narg = 0;
     msg->integer = 0;
 
     msg->err = 0;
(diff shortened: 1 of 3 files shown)


TysonAndre:
A different approach based on https://github.com/twitter/twemproxy/pull/565#issuecomment-871854468 by qingping209 was merged

The above links were to the wrong PR (only saw the most recent "multi-bulk" when looking for what to link to that commit, didn't think there were more than one PR for that functionality at the time) - that should have been linked to 565

I'd guess that properly supporting sharded/fragmented commands required a bit more state tracking than what's proposed in your PR but only skimmed it after merging 612 - e.g. if a fragmented multi-bulk command is sent to two servers with responses of length 3 and 5, then twemproxy must also track the top-level length of 3+5=8 to generate the correct response array length header when combining fragments).

- There might have been other array types - it's been a while since the commits I made to the fork https://github.com/twitter/twemproxy/pull/612 was based on a subset of
- I'd based my approach on #565 because the CLA had been signed

PR #589, opened 2020-07-16
redis: add support for unlink

kornrunner:
`UNLINK` command for redis is currently not implemented.
This change implements missing functionality.


src/proto/nc_redis.c
```diff
@@ -2787,6 +2807,9 @@ redis_post_coalesce(struct msg *r)
     case MSG_REQ_REDIS_DEL:
         return redis_post_coalesce_del(r);
 
+    case MSG_REQ_REDIS_UNLINK:
+        return redis_post_coalesce_unlink(r);
```
TysonAndre, line 2811: I think this could just call redis_post_coalesce_del if the method implementations  are the same (e.g. add a comment that the method is used for DEL or UNLINK)

src/proto/nc_redis.c
```diff
@@ -2380,9 +2386,6 @@ redis_pre_coalesce(struct msg *r)
 
     switch (r->type) {
     case MSG_RSP_REDIS_INTEGER:
-        /* only redis 'del' fragmented request sends back integer reply */
-        ASSERT(pr->type == MSG_REQ_REDIS_DEL);
```
TysonAndre, line 2384: `ASSERT(pr->type == MSG_REQ_REDIS_DEL || pr->type == MSG_REQ_REDIS_UNLINK);` would also work

TysonAndre (commented):
Not a maintainer but moderately familiar with twemproxy - it seems like it'd work but I had minor comments based on what I'd seen elsewhere

TysonAndre:
Support for unlink was added as part of a larger change to add support for many new/updated redis commands, see 1239ef9e2f14f954b0f5da22369c0d994a347e45 and https://github.com/twitter/twemproxy/pull/599 (etc.)

(I was added as a collaborator this week)

PR #590, opened 2020-09-18
Abort connection on error

katsew:
Problem

If the backend server of twemproxy down, twemproxy sends an error message like `-ERR Host is down`.
However, some of the client library ( in our case phpredis ) cannot recognize it as an error but ignores it.

Solution

If the error occurred on the backend of twemproxy, twemproxy sends RST packet back to the client,
so that the client can raise an exception correctly.

Result

The client library (phpredis) now throws an exception when redis, which is our backend of twemproxy, is down.


katsew:
Sorry, wrong repo to send PR.

Issue #563, opened 2018-08-26
Segmentation-Fault: ncontinuum versus Pool Server-Count


Sha0:
This Is Not A Contribution:

Good day.

I've encountered reproducible segmentation-faults while using `auto_eject_hosts=true`.  Debugging led me to an out-of-bounds attempted access of a `sub_msgs` array.

In these functions:
- `src/proto/nc_memcache.c` : `memcache_fragment_retrieval`
- `src/proto/nc_redis.c` : `redis_fragment_argx`

We find the following:

`  sub_msgs = nc_zalloc(ncontinuum * sizeof(*sub_msgs));`

Later, we find this:

`  uint32_t idx = msg_backend_idx(r, kpos->start, kpos->end - kpos->start);`

But if I've followed the code-base correctly, the back-end index is an index into a pool's total servers (both live and dead), whereas `ncontinuum` is the number of populated slots in the pool's continuum.

Would you be kind enough to help me to understand the intention behind this apparent mismatch, or is it a bug?  Is it the case that we actually wish for that `zalloc` to use the pool's total number of servers, instead?  If so, that seems to complicate the usefulness of passing `ncontinuum` from above.

The current code works when `ncontinuum` >= the total number of servers in the pool, but using `auto_eject_hosts=true` and dropping servers from the continuum, this condition doesn't always hold.

Thank you for your time.

TysonAndre:
(I'm not one of the maintainers of twemproxy - I have a understanding of bits and pieces of twemproxy from working on integrating a complicated custom patch into twemproxy) 

I ran into the same question in a nutcracker setup having auto_eject_hosts=true (and server_connections=16, which is unrelated)

Possibly related to https://github.com/twitter/twemproxy/issues/481 since the stack trace is similar


I'm also concerned about whether it's possible for memcache_fragment to be called with `ncontinuum == 0` `nc_zalloc(ncontinuum * sizeof(*sub_msgs))` - `idx = 0`, but would be pointing to invalid memory if sub_msgs was an array of size 0 bytes (which is part of the general case)

```asm
[2021-02-23 14:08:21.702] nc_server.c:589 connected on s 399 to server '10.50.24.30:11211:1'
nutcracker(nc_stacktrace_fd+0x17)[0x418677]
nutcracker(signal_handler+0xa9)[0x416339]
/lib64/libpthread.so.0[0x3dc720f7e0]
nutcracker(memcache_fragment+0xf1)[0x41c821]
[...]

PR #595, opened 2021-02-24
Use nservers instead of ncontinuum to fix segfault when all hosts are ejected

TysonAndre:
ncontinuum is `160 * the number of **non-ejected** hosts` for the ketama
distribution, where 160 is the ketama consistent hashing points per server.
However, the index returned by msg_backend_idx is a number between
0..nservers-1.

- This is a waste of time zeroing out 160 times as much memory as needed
- When there are 0 servers, this previously malloc()ed a pointer of size 0 bytes
  and nutcracker would write to invalid memory.

Also, update redis nosetest expectation because newer redis versions changed the
error message to mention replica instead.

Also, update pip instructions to something supporting the latest python 3 versions (pip dropped python 2 support recently, ubuntu 20.04 doesn't include pip for python2 and https://bootstrap.pypa.io/2.7/get-pip.py` has to be downloaded and run)

Fixes https://github.com/twitter/twemproxy/issues/563 for redis and memcached.
See the discussion there for more details.

Problem

nutcracker segfaults or has memory corruption when a memcache/redis request is issued when all hosts are ejected with auto_eject_hosts: true (e.g. during a networking outage)
See https://github.com/twitter/twemproxy/issues/563

Solution

Fix the corruption by allocating an array of the correct size (nservers instead of ncontinuum)

Also, don't fragment memcache get commands with a single key

Fragmentation would likely be slower.
It would involve splitting up the request into multiple smaller
requests (allocating and zeroing out an array)
and concatenating strings to rebuild the original request.

Fragmentation is also more likely to be prone to edge cases such as
those mentioned in this PR
and #595

Result

Fix segfaults and memory corruption


TysonAndre:
could @manjuraj or another maintainer take a look? I confirmed locally that idx was less than nservers and documentation for redis indicates that the parameter was intended to be the number of servers.

Travis CI was already failing for unrelated reasons (e.g. wrong command to install socat in current ubuntu images) but tests pass locally

TysonAndre:
This patch was deployed yesterday in an application I'm working on on a large number of nutcracker instances and php servers (using a build of nutcracker which already had unrelated patches)

Average nutcracker latency went down by roughly 0.01 milliseconds(10 microseconds), e.g. from 0.21 milliseconds to 0.20 milliseconds (because 160 times less pointers are being zeroed out) - individual nutcracker instances were at around 25% cpu before/after this change.

No new nutcracker issues were caused by this patch. The networking issues that caused memcache hosts to be ejected as a symptom may have been less severe today.

TysonAndre:
> This patch was deployed yesterday in an application I'm working on on a large number of nutcracker instances and php servers (using a build of nutcracker which already had unrelated patches)

The application continues to be stable despite occasional host evictions, and the reason for host evictions was identified.

PR #598, opened 2021-04-09
Fix failing unit tests, run in Docker

TysonAndre:
Problem

Travis tests are currently failing

Solution

Running it in docker makes it easier to run this in other CI providers
and to easily reproduce any failures seen in Travis locally.

Move to python 3

Rewrite travis.sh from scratch

Result

Travis tests will pass, and it will be easier to run them locally if python 3 is used instead of 2


TysonAndre:
Oh, right, I'd forgotten about https://travis-ci.community/t/docker-hub-you-have-reached-your-pull-rate-limit/10517 and https://blog.travis-ci.com/docker-rate-limits due to recent changes in docker hub. I thought there would have been caching.

I think other providers would work, but I'm not familiar with github actions and other workflows would require owners to enable the integration

TysonAndre:
It looks like actions would need to be enabled by the owners of this repository for the latest commit to work?

I validated that this would work in a fork: https://github.com/ifwe/twemproxy/pull/30

TysonAndre:
It's been a while since the last time I set up GitHub actions through a Pull Request.

Apparently, they'll only get triggered if github actions is already configured for the target branch - it's already triggered in https://github.com/twitter/twemproxy/actions/runs/986468738

PR #611, opened 2021-07-01
Fix parsing of redis error message without space

### LINKED ISSUE #404: 0.4.1 Response corruption with timeout and lua script
We've found an issue whereby Twemproxy seems to get into a situation where responses are corrupted and mixed incorrectly between clients. The issue seems to be triggered by evaluation of a lua script combined with a response timeout. After a request is times out (and the client receives an "-ERR Connection timed out" Redis response), subsequent requests can see partial responses from other clients requests.

The minimal example I've been able to create that exhibits the issue:

``` lua
    local id_counter_key = KEYS[1]
    local x_key = KEYS[2]
    local id = -1
    local id_key = ""
    local x = ARGV[1]
    if redis.call('exists', x_key) == 1 then
        return {err="10001"}
    end
    id = redis.call('incr', id_counter_key)
    id_key = 'test_data:' .. id
    redis.call('hset', id_key, 'id', id)
    redis.call('hset', id_key, 'x', x)
    redis.call('set', x_key, id)
    return id
```

Combined with a configuration like:

``` yaml
user:
  listen: 0.0.0.0:22120
  redis: true
  hash: fnv1a_64a
  distribution: ketama
  auto_eject_hosts: false
  timeout: 5000
  server_retry_timeout: 1000
  server_failure_limit: 1 
  servers:
   - redis.test.com:6379:1 server
```

The problem appears using eval or evalsha, I can't reproduce it without a lua script that times out. It is sufficient to force timeouts by setting the twemproxy pool timeout value sufficiently low (e.g. 10ms) to see the issue, but the issue has been seen with 'real' response timeouts of both 5 and 20 secs.

I am also only able to reproduce this when 2 cases hold:
- The script has a error-condition exit such as the one above if the key exists;
- And also there is a separate process making set and get commands through twemproxy that have data of the form "nn:nn:nn" e.g. "12:34:56".

A minimal example of that 'second process':

``` sh
while true
do
  echo
  date
  nslookup $ENDPOINT | grep Address: | tail -n+2
  echo -e "get test_counter\nset test `date +%H:%M:%S`\nget test" \
    | redis-cli -h $ENDPOINT -p $PORT

  sleep 1
done
```

I have been trying to get a minimal example together but struggling to reproduce the situation when there isn't a secondary client, but I will keep working on this on Monday to get to the bottom of this, but thought it was worth reporting now due to the severity.

Here is a Python script that should demonstrate the issue when run in conjunction with the above shell loop:

``` python
from multiprocessing import Process
import random
import signal
import sys
import time

from redis import StrictRedis, ResponseError


WORKER_DELAY_SECS = 0.0001
NUM_WORKERS = 10


SCRIPT = """
    local id_counter_key = KEYS[1]
    local x_key = KEYS[2]
    local id = -1
    local id_key = ""
    local x = ARGV[1]
    if redis.call('exists', x_key) == 1 then
        return {err="10001"}
    end
    id = redis.call('incr', id_counter_key)
    id_key = 'test_data:' .. id
    redis.call('hset', id_key, 'id', id)
    redis.call('hset', id_key, 'x', x)
    redis.call('set', x_key, id)
    return id
"""


def get_twemproxy_redis():
    return StrictRedis("localhost", 22120)


def worker():
    user_redis = get_twemproxy_redis()

    while True:
        try:
            x = random.randint(1, 1000)

            result = user_redis.eval(SCRIPT, 2, "test_counter_key", "test_data_key", x)
            int(result)

        except ResponseError as e:
            pass

        time.sleep(WORKER_DELAY_SECS)


def create_data():
    subprocesses = []
    for i in range(NUM_WORKERS):
        p = Process(target=worker, name="worker-{}".format(i))
        subprocesses.append(p)

    for p in subprocesses:
        p.start()


if __name__ == "__main__":
    create_data()
```

Within a few seconds of both scripts running:

```
(venv)root@app-test:/opt/app# python run_twemproxy_soak_test.py 
Process worker-9:
Traceback (most recent call last):
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "run_twemproxy_soak_test.py", line 44, in worker
    int(result)
ValueError: invalid literal for int() with base 10: '-ERR Inv'
```

TysonAndre:
And add unit tests of redis messages being parsed successfully

Add testcase of error with no space

The test case was cherry-picked from https://github.com/twitter/twemproxy/pull/406

The alternate fix a74db76ce816caa14e8a85548f34088f9c594e6e was used - both do the same thing, not subtracting is personal preference

Amended to work with the updated test framework

The implementation/C unit test was amended to follow twemproxy's c style guide

Closes #404

Solution

Properly find the end of the line for a redis error when there's no space after the error type

Result

Redis errors that are just the error type will be processed by twemproxy correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,24 @@
+ 2021-??-??  Tyson Andre
+    * twemproxy: version 0.4.2 release (dev)
+	  Fix parsing of redis error response for error type with no space,
+	  add tests (tyson, tom dalton)
+	  Update integration tests, add C unit test suite for 'make check' (tyson)
+	  Increase the maximum host length+port+identifier to 273
+	  in ketama_update (李广博)
+	  Always initialize file permissions field when listening on a unix domain
+	  socket (tyson)
+	  Use number of servers instead of number of points on the continuum when
+	  sharding requests to backend services to improve sharding performance
+	  and fix potential invalid memory access when all hosts were ejected
+	  from a pool. (tyson)
+	  Don't fragment memcache/redis get commands when they only have a single
+	  key (improves performance and error handling of single key case) (tyson)
+	  Don't let requests hang when there is a dns error when processing a
+	  fragmented request (e.g. multiget with multiple keys) (tyson)
+	  Allow extra parameters for redis spop (charsyam)
+	  Update documentation and README (various)
+	  Fix memory leak bug for redis mset (deep011)
+
  2015-22-06  Manju Rajashekhar  <manj@cs.stanford.edu>
     * twemproxy: version 0.4.1 release
       backend server hostnames are resolved lazily
--- a/src/proto/nc_redis.c
+++ b/src/proto/nc_redis.c
@@ -1920,7 +1920,12 @@ redis_parse_rsp(struct msg *r)
 
                     break;
                 }
-                state = SW_RUNTO_CRLF;
+                if (ch == '\r') {
+                    state = SW_ALMOST_DONE;
+                } else {
+                    /* Read remaining characters until '\r' */
+                    state = SW_RUNTO_CRLF;
+                }
             }
 
             break;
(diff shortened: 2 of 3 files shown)


TysonAndre:
The original work on https://github.com/twitter/twemproxy/pull/406 seems to predate the addition of the CLAAssistant bot, so omitting the integration test amended for the updated integration test framework for now, the C unit test should cover it.

(The integration test and alternate fix approach was merged in and later reverted years ago when investigating something else, so this is probably excessive, I'll add the integration test if the CLA is signed)

PR #660, opened 2021-12-09
Reduce one syscal for fcntl

theidexisted:
According to the man page: https://man7.org/linux/man-pages/man2/socket.2.html
       Since Linux 2.6.27, the type argument serves a second purpose: in
       addition to specifying a socket type, it may include the bitwise
       OR of any of the following values, to modify the behavior of
       socket():

       SOCK_NONBLOCK
              Set the O_NONBLOCK file status flag on the open file
              description (see open(2)) referred to by the new file
              descriptor.  Using this flag saves extra calls to fcntl(2)
              to achieve the same result.

Problem

Explain the context and why you're making that change.  What is the
problem you're trying to solve? In some cases there is not a problem
and this can be thought of being the motivation for your change.

Solution

Describe the modifications you've done.

Result

What will change as a result of your pull request? Note that sometimes
this section is unnecessary because it is self-explanatory based on
the solution.

### DIFF
--- a/src/nc_proxy.c
+++ b/src/nc_proxy.c
@@ -128,7 +128,7 @@ proxy_listen(struct context *ctx, struct conn *p)
 
     ASSERT(p->proxy);
 
-    p->sd = socket(p->family, SOCK_STREAM, 0);
+    p->sd = socket(p->family, SOCK_STREAM | SOCK_NONBLOCK, 0);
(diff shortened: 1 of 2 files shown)


yuerugou54:
马晓栋已经收到您的来件，谢谢!

PR #676, opened 2022-12-23
add redis script command support

lukemakeit:
Problem

redis script command not support.

Solution

Send the script command to all Redis servers for execution, and get one of the results and return it to the user client.

Result

```
> script load "return redis.call('hset',KEYS[1],KEYS[1],KEYS[1])"
"dbbae75a09f1390aaf069fb60e951ec23cab7a15"
> script exists dbbae75a09f1390aaf069fb60e951ec23cab7a15
1) (integer) 1
```


yuerugou54:
马晓栋已经收到您的来件，谢谢!

PR #677, opened 2023-01-04
redis script command support

lukemakeit:
Problem

redis script command not support.

Solution

Send the script command to all Redis servers for execution, and get one of the results and return it to the user client.

Result

```
> script load "return redis.call('hset',KEYS[1],KEYS[1],KEYS[1])"
"dbbae75a09f1390aaf069fb60e951ec23cab7a15"
> script exists dbbae75a09f1390aaf069fb60e951ec23cab7a15
1) (integer) 1
```

### DIFF
--- a/src/nc_server.c
+++ b/src/nc_server.c
@@ -700,12 +700,16 @@ server_pool_idx(const struct server_pool *pool, const uint8_t *key, uint32_t key
 }
 
 static struct server *
-server_pool_server(struct server_pool *pool, const uint8_t *key, uint32_t keylen)
+server_pool_server(struct server_pool *pool, struct msg *r, const uint8_t *key, uint32_t keylen)
 {
     struct server *server;
     uint32_t idx;
 
-    idx = server_pool_idx(pool, key, keylen);
+    if (r->type == MSG_REQ_REDIS_SCRIPT) {
+        idx = r->redis_script_idx;
+    } else {
+        idx = server_pool_idx(pool, key, keylen);
+    }
     server = array_get(&pool->server, idx);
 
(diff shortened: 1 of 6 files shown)


src/nc_server.c
```diff
@@ -700,12 +700,16 @@ server_pool_idx(const struct server_pool *pool, const uint8_t *key, uint32_t key
...
 static struct server *
-server_pool_server(struct server_pool *pool, const uint8_t *key, uint32_t keylen)
+server_pool_server(struct server_pool *pool, struct msg *r, const uint8_t *key, uint32_t keylen)
 {
     struct server *server;
     uint32_t idx;
 
-    idx = server_pool_idx(pool, key, keylen);
+    if (r->type == MSG_REQ_REDIS_SCRIPT) {
+        idx = r->redis_script_idx;
```
w4096, line 709: Add a member named `server_index` in `struct msg`.  If the value is >=0, use `msg->server_index` as target server index. In this way, we don't need `redis_script_idx` and `scan_server_idx` at all.

src/proto/nc_redis.c
```diff
@@ -2945,10 +2982,97 @@ redis_fragment_argx(struct msg *r, uint32_t nserver, struct msg_tqh *frag_msgq,
     return NC_OK;
 }
 
+static rstatus_t redis_fragment_script(struct msg *r, struct msg_tqh *frag_msgq) {
```
w4096, line 2985: If we can implement a function `redis_fragment_broadcast` which can broadcast a request to all servers, we can implement broadcast commands like `DBSIZE`, `FLUSHALL` easily. What I mean is this function is not general enough, which can only handle `SCRIPT` command.

PR #678, opened 2023-01-04
redis scan command support

lukemakeit:
Problem

redis scan command not support.

Solution

image

Note:
`server_idx` is saved by the lowest 12 bits of cursor, so the number of servers of the twemproxy must be less than 4096.

Result

```
1.1.1.1:60100> scan 0
1) "1"
2) 1) "d"
   2) "b"
   3) "c"
   4) "a"
1.1.1.1:60100> scan 1
1) "0"
2) 1) "a"
   2) "b"
```


src/nc_server.c
```diff
@@ -700,12 +700,37 @@ server_pool_idx(const struct server_pool *pool, const uint8_t *key, uint32_t key
...
+            real_cursor = (cursor >> NC_MAX_NSERVER_BITS);
+            sprintf(format,"%%0%dd",keylen);
+            sprintf(arr,format,real_cursor);
+            nc_memcpy(key,arr,keylen);
+        }
+        r->scan_server_idx=idx;
+        r->max_server_idx=(pool->server).nelem;
+    }else{
+        idx = server_pool_idx(pool, key, keylen);
+    }
```
w4096, line 714: Change the cursor of request in `req_filter` and save the `server_index` on msg in the mean time.

src/proto/nc_redis.c
```diff
@@ -3053,6 +3069,46 @@ redis_post_coalesce_mget(struct msg *request)
     }
 }
 
+void redis_post_coalesce_scan(struct msg *request) {
```
w4096: move those code in `rsp_filter`

src/proto/nc_redis.c
```diff
@@ -2963,6 +2976,9 @@ redis_fragment(struct msg *r, uint32_t nserver, struct msg_tqh *frag_msgq)
     case MSG_REQ_REDIS_MSET:
         return redis_fragment_argx(r, nserver, frag_msgq, 2);
 
+    case MSG_REQ_REDIS_SCAN:
+        return redis_fragment_scan(r,frag_msgq);
```
w4096: The scan command doesn't need fragment, the thing you want do is change the cursor, you can do this in `req_filter`.
When response received you need change the cursor again, you can do this in `rsp_filter`.

src/nc_server.h
```diff
@@ -137,7 +137,7 @@ void server_connected(struct context *ctx, struct conn *conn);
 void server_ok(struct context *ctx, struct conn *conn);
 
 uint32_t server_pool_idx(const struct server_pool *pool, const uint8_t *key, uint32_t keylen);
-struct conn *server_pool_conn(struct context *ctx, struct server_pool *pool, const uint8_t *key, uint32_t keylen);
+struct conn *server_pool_conn(struct context *ctx, struct server_pool *pool, const uint8_t *key, uint32_t keylen, struct msg *msg);
```
w4096: I think put `msg` in front of key and keylen is better.

src/nc_server.c
```diff
@@ -700,12 +700,37 @@ server_pool_idx(const struct server_pool *pool, const uint8_t *key, uint32_t key
...
+            */
+            cursor=strtoull(key,NULL,10);
+            idx = cursor & NC_MAX_NSERVER_MASK;
+            real_cursor = (cursor >> NC_MAX_NSERVER_BITS);
+            sprintf(format,"%%0%dd",keylen);
+            sprintf(arr,format,real_cursor);
+            nc_memcpy(key,arr,keylen);
+        }
+        r->scan_server_idx=idx;
+        r->max_server_idx=(pool->server).nelem;
```
w4096: I think `max_server_idx` is unnecessary, since you can always known the number of server by `array_n(&pool->server)` .

lukemakeit:
Hi @wy-ei , I have re-committed according to your suggestion, thanks.

src/nc_request.c
```diff
@@ -511,6 +520,31 @@ req_filter(struct conn *conn, struct msg *msg)
         msg->noforward = 1;
     }
 
+    if (msg->type == MSG_REQ_REDIS_SCAN) {
+        ASSERT(array_n(msg->keys) > 0);
```
w4096: move this into a function and call the function here like what you did in `rsp_filter`.

src/nc_message.h
```diff
@@ -300,6 +301,8 @@ struct msg {
     unsigned             fdone:1;         /* all fragments are done? */
     unsigned             swallow:1;       /* swallow response? */
     unsigned             redis:1;         /* redis? */
+
+    uint32_t             server_index; /* used for store the redis server index in server pool */
 };
```
w4096, line 308: this comment is not clear enough. how about this one:

the server index which the requstion should be forwarded.

w4096:
@lukexwang 

I have see the change you made. it's better. And I think the code can be optimized again.  I have leave some more suggestions.

src/nc_message.h
lukexwang: ok

lukemakeit: > this comment is not clear enough. how about this one:
> 
> the server index which the requstion should be forwarded.

ok

src/nc_request.c
lukemakeit: Good advice

lukemakeit:
Hi @wy-ei  , thank you for your suggestion. I have re-committed.

lukemakeit:
Hello @wy-ei , 空了辛苦看下哈 ^_^

yuerugou54:
马晓栋已经收到您的来件，谢谢!

PR #684, opened 2024-02-28
Add select 0 support

lukemakeit:
Problem

add select 0 support

Result

```
127.0.0.1:60100> select 0
OK
127.0.0.1:60100> select 1
(error) ERR only select 0 support
```


yuerugou54:
马晓栋已经收到您的来件，谢谢!

